Issue #66, opened 2015-05-06
custom dataTablesJSON


vnijs:
I can see how to create my own version of dataTablesJSON but I'm not sure how to get DT to use it. Could dataTableAjax take a function as an argument?

A related question if about ranges. dataTablesJSON seems to use commaToRange but the filter boxes use '...' instead. Could you explain how this work?

My question is related to https://github.com/rstudio/DT/issues/50
@kismsu Could you perhaps share your example?

yihui:
`...` is just used for the presentation purpose; internally the two values of the range are joined by `,` before they are sent to the server: https://github.com/rstudio/DT/blob/9f5f0e5478ac241b7cd8361412a96299821f144e/inst/htmlwidgets/datatables.js#L203

And yes, I can add a function argument to `dataTableAjax()` that defaults to `dataTablesJSON` if you want.

vnijs:
Ah. I seem to recall that both `, 5` and `5, ` used to work in shiny::datatables. `... 5` works but not `5 ...`. To be honest, I would probably prefer using `,`. There is not a lot of space in these boxes when you have 10+ columns :)

By the way, there is no option to hide the `remove` glyph at this time correct?

If there is an alternative way to get DT to use 'my' function that is fine. I just couldn't figure out how to do it. If you are ok with changing `dataTableAjax` that works for me. 

Thanks Yihui!

yihui:
You are right. I have not addressed the issue of the crowded search boxes yet.

For `,5` and `5,`, I guess they are no longer necessary in DT, because in these cases, the missing end will be automatically replaced by the min or max. (If it is not, it must be a bug)

vnijs:
How does DT know what end is missing? It still needs ... (or , ) right? If so, then , might be preferable.

`... 5000` is converted to `338.00 ... 5000.00` but `5000 ...` is converted to `5000.00 ... 5000.00`. You can try (see) at link below:

http://vnijs.rady.ucsd.edu:3838/marketing/?SSUID=36a8d30c0e175c00b4f3622ac4e82490

yihui:
That is a bug. Should be fixed now.

vnijs:
Works! Are you open to changing the `...` to `,` and hiding the `remove` glyph? If so do you prefer a pull request?

yihui:
[...]

Issue #49, opened 2015-04-16
Column filter css affected when switching to tab/page with shiny slider


vnijs:
In style.css I have:

table.dataTable tfoot td, table.dataTable thead td {
  padding-top: 0px;
  padding-bottom: 0px;
  padding-left: 0px;
  padding-right: 0px;
}

In Data > View this looks fine. Go to Base > Single mean (which has a filter) and then back to Data > View and you will see the css has been changed. This happens for any page with a slider (e.g, all tools in Base except Cross-tabs, Regression (OLS and GLM, but not Correlation which doesn't have a slider). To test this some more I added a slider to the Data > Visualize tab and the same effect occurred. Switching back and forth changes the CSS.

Note that this effect also occurs when I comment out the CSS listed above. In that case the default CSS for the column headers/footers is still changed when switching back and forth to pages with shiny sliders.

See for link below for example:

https://vnijs.shinyapps.io/quant

All package versions used are at: https://github.com/vnijs/radiant_miniCRAN

yihui:
Can you provide two screenshots that show how the styles change from one case to the other? I could see some subtle padding differences, and that does not seem to be too bad in my eyes. Perhaps you meant something more significant that I failed to see.

vnijs:
See attached. Certainly not horrible or anything. Just thought the interaction with shiny sliders was worth mentioning.

![screen shot 2015-05-30 at 3 23 20 pm](https://cloud.githubusercontent.com/assets/2998956/7899637/ff0aed7c-06df-11e5-8c73-6e1a74ba7a83.png)

![screen shot 2015-05-30 at 3 23 52 pm](https://cloud.githubusercontent.com/assets/2998956/7899638/0645fbcc-06e0-11e5-8657-23af493d49dd.png)

yihui:
Okay, that is worse than I imagined, but given that you can get rid of the clear buttons now, I guess this is no longer an urgent issue.

vnijs:
Just an FYI ... without the clear button the text from different cells can overlap. See screenshot. Like I said, not a big deal but enough for me to prefer `plain = TRUE` for now. Thanks for that option!

![screen shot 2015-05-30 at 3 40 02 pm](https://cloud.githubusercontent.com/assets/2998956/7899682/6fa4399c-06e2-11e5-8868-23702ac04ef0.png)

yihui:
Ugh... :dizzy_face:

PR #70, opened 2015-05-09
Column filter spacing + custom dataTablesJSON function

vnijs:
@yihui I tried a few things to see if I could increase the spacing in the column filter boxes. For some reason the padding on the right is currently really high. I removed the form-control class in a few places and, even though the look is not that stylish, the spacing is much improved. With these changes issue #49 also does not occur which is a nice side-effect. I actually like the remove-circle below the box. See the Data > View tab at http://vnijs.rady.ucsd.edu:3838/base (Data > View tab) for the alternative 'look'

I also added an argument to `dataTableAjax` to make it easier to create a custom `dataTablesJSON` function - #66.

Any suggestions on how to make the sliders a bit coarser? It is currently very hard to get to a whole number which is what a user would most likely want, especially for integers.


yihui:
Thanks! The new argument for `dataTablesAjax()` is fine. I need to think more about what to do with the clear button and the step length of the slider.

BTW, if you have not signed the CLA for RStudio before, here is what you need to do before we can accept your pull requests to RStudio's Github repositories: https://github.com/rstudio/shiny/blob/master/CONTRIBUTING.md

vnijs:
Great. Just signed and sent the CLA.

vnijs:
FYI Just got the confirmation email from JJ.

vnijs:
FYI Just got the confirmation email from JJ

yihui:
Okay, could you separate the two things? Ideally let's do one thing per PR, so we can accept one without waiting for the other being revised. As I said, the `dataTableAjax` argument looks fine. Perhaps a name `filter` (or `filterFun`) sounds better than `toJSONfun`.

I'd strongly recommend you to use GIT branches for pull requests in the future, unless you are fairly confident it is a trivial change and will definitely be accepted, otherwise chances are you will have to revise, rebase, force push, and resubmit, which are not good for the master branch. The second commit in this PR could have been avoided by rebasing (cf https://github.com/yihui/knitr/blob/master/CONTRIBUTING.md#pull-requests).

Thanks!

vnijs:
Thanks for the suggestions. This is such a minor change that it would be a lot easier if you add the argument in master. For the filter boxes, removing `form-control` is probably not a great way to implement what I had in mind. When using DT in a shiny app the square boxes (also on `search`) are a bit ugly and inconsistent with the rest of the formatting. I was hoping that this approach might help a bit for you to come up with a better solution :)

yihui:
Please see http://rstudio.github.io/DT/008-filter.html

vnijs:
I was just looking at this. Thanks

vnijs:
Really nice enhancements @yihui. The sliders for integer values work great. Thank you!

I haven't had a chance to try-out customizing the server-side filtering etc. but for now there are just two remaining items on my DT _wish-list_. 1. have stateSave working for column filters and 2. css fix for slider location in wide tables (see screen shot). For 2, as you can see, the slider is far from the active cell in these cases. This does not happen for factors by the way.

screen shot 2015-05-30 at 3 36 14 pm

yihui:
I'll work on 1 this evening. I'll have to think more about 2.

vnijs:
Thanks. FYI I just noticed something that happens for numbers with > 2 decimal places. As you can see in the 1st screenshot the slider shows all (?) decimals. Then when you use the slider (2nd screenshot) the numbers are formatted to 2 decimal places. Perhaps this is what you want if you are looking for min and max although in the 2nd screenshot max is also formatted to 2 decimals but was not changed. Using 2 decimals in the initial slider might be (slightly) preferred.

screen shot 2015-05-30 at 4 00 22 pm

screen shot 2015-05-30 at 4 00 36 pm

yihui:
Could you try to file a new issue as you find new problems? Otherwise I feel I can never close an issue since it often quickly diverges to several different issues, and the size of the bullet points just keeps on growing in the response...

For your wish list:

1. I thought a while and sorry I don't have a good solution yet.
2. Same as #73. Fixed. Again, please try not to blend different issues into one thread.

For this particular case, I guess it does not hurt much, but I discovered a much worse bug: when all numbers are smaller than 0.001, the range will be rounded to `[0,0]`, which will fail the slider...

vnijs:
Will do. Thanks

PR #84, opened 2015-06-01
Allow data frames to be used as renderDataTable values

jcheng5:


R/shiny.R
```diff
@@ -53,7 +53,18 @@ renderDataTable = function(expr, env = parent.frame(), quoted = FALSE, ...) {
...
-  htmlwidgets::shinyRenderWidget(expr, dataTableOutput, env, quoted = TRUE)
+
+  exprFunc <- shiny::exprToFunction(expr, env, quoted = TRUE)
+  widgetFunc <- function() {
+    x <- exprFunc()
+    if (is.data.frame(x)) {
+      datatable(x)
+    } else {
+      x
+    }
```
jcheng5: Actually I think we could use this spot in the code to implement `server=TRUE` without requiring help from htmlwidgets. This is effectively the same thing as a Shiny-only preRenderHook, and the name and session are both available I believe.

jcheng5: Oh, it's trickier than that (but still possible). I'll put together a separate PR to demonstrate.

yihui:
That is great. Thanks!

PR #86, opened 2015-06-01
Allow data frames to be used as renderDataTable values

jcheng5:
Sorry, last time I had the final `quoted=TRUE` where it should've been `quoted=FALSE`.


yihui:
No worries. I tested #84 before I merged it, and it worked.

PR #87, opened 2015-06-01
Simplify server handling

jcheng5:
This is a **straw man** proposal for how we could do this without requiring https://github.com/ramnathv/htmlwidgets/pull/122. It pulls in your no-register branch.

There are two hacks to point out, plus an open issue I think we should discuss.

Hack 1 is that the `preRenderHook` is evaluated and then removed in `widgetFunc`. This works fine but it feels like it definitely violates the spirit of `preRenderHook`. My first attempt tried to remove the preRenderHook altogether and copy that logic into the Shiny render function, but I couldn't figure out how to get the `escape` and `colnames` out. I thought that would be a pretty easy refactor for you, but I didn't want to go that far afield for this proof-of-concept.

Hack 2 is the usage of `currentSession` and `currentOutputName` variables, that we use to transmit the session and name from the render function to the `widgetFunc`. This is a bit inelegant but I'm totally fine with shipping with this, as it's unlikely to cause bugs and certainly doesn't affect the user experience negatively.

Open issue is whether `server = TRUE` belongs on `datatable()` in the first place, or whether it should be on `renderDataTable` as it is in Shiny today. I think the latter makes at least as much sense as the former, since `server` mode is only possible with Shiny anyway. On the other hand, it would be unfortunate to break DT backwards compatibility. On the gripping hand, I'd rather break compatibility with DT now than in the future, and I'd slightly prefer to break with DT than with the renderDataTable function Shiny has shipped for ages now.

I'll be at lunch from 12pm-1:20pm PDT or so, but I'll be available to talk about this after that if you want.

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -147,7 +147,9 @@ datatable = function(
   if (server) {
     origData = data
     data = NULL
-    options = fixServerOptions(options, escape, colnames)
+    # indices of columns that need to be escaped
+    attr(options, 'escapeIdx') = escapeToConfig(escape, colnames)
+    options = fixServerOptions(options)
 
     # register the data object in a shiny session
     registerData = function(instance, shinysession, name) {
@@ -159,7 +161,7 @@ datatable = function(
 
       url = sessionDataURL(shinysession, origData, name, dataTablesFilter)
       options$ajax$url = url
-      instance$x$options = fixServerOptions(options, escape, colnames)
+      instance$x$options = fixServerOptions(options)
 
       instance
     }
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -74,7 +74,7 @@ renderDataTable = function(expr, env = parent.frame(), quoted = FALSE, ...) {
 
   renderFunc <- htmlwidgets::shinyRenderWidget(widgetFunc(), dataTableOutput, environment(), FALSE)
 
-  markRenderFunction(dataTableOutput, function(shinysession, name, ...) {
+  shiny::markRenderFunction(dataTableOutput, function(shinysession, name, ...) {
     currentSession <<- shinysession
     currentOutputName <<- name
     on.exit({
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -53,7 +53,18 @@ renderDataTable = function(expr, env = parent.frame(), quoted = FALSE, ...) {
   )
   checkShinyVersion()
   if (!quoted) expr = substitute(expr)
-  htmlwidgets::shinyRenderWidget(expr, dataTableOutput, env, quoted = TRUE)
+
+  exprFunc <- shiny::exprToFunction(expr, env, quoted = TRUE)
+  widgetFunc <- function() {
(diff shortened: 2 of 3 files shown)


jcheng5:
Example:

```r
library(shiny)
library(DT)
library(ggplot2)

ui <- fluidPage(
  DT::dataTableOutput("dt")
)

  dt <- reactive(diamonds)
  output$dt <- DT::renderDataTable({
    DT::datatable(dt(), server = TRUE)
  })
}

shinyApp(ui, server)
```

jcheng5:
Proposed syntax (mirrors deprecated Shiny renderDataTable):

```r
library(shiny)
library(DT)
library(ggplot2)

ui <- fluidPage(
  DT::dataTableOutput("dt")
)

  dt <- reactive(diamonds)
  output$dt <- DT::renderDataTable({
    dt()
  }, server = TRUE)
}

shinyApp(ui, server)
```

yihui:
1. I'm still thinking about Hack 1.
2. Hack 2 is totally fine to me.

Re: the open issue, what I was thinking originally was actually to pass `...` from `renderDataTable()` to `datatable()` when `expr` is a data frame, so shiny users can be more comfortable with switching from shiny to DT, e.g. both `server = TRUE` and `escape = TRUE` can be passed from `renderDataTable()` to `datatable()`.

jcheng5:
I agree passing `...` to `datatable()` makes the transition easier. Is it where we want to end up long-term?

For `server` specifically, do you think that belongs on the datatable, or does it belong on the renderDataTable, period? I don't necessarily think of that as a property of the widget itself, but as a rendering decision. Does that make sense?

jcheng5:
To be clear, with the latter question, I'm asking whether even in the case of returning a datatable widget from the `renderDataTable` expression, you should still pass `server=TRUE` as an argument to `renderDataTable` and not to `datatable`.

yihui:
When server=TRUE has been specified in datatable() to be passed to renderDataTable(), there is no need to pass server=TRUE to renderDataTable() again (if you do, it will just be silently ignored).

It was important to have server=TRUE as the argument of datatable() because I wanted to know when to discard the original data (we don't want to encode a big data object in JSON): https://github.com/rstudio/DT/blob/aa492c439c0b43d1734ef4bce854a551f9ecfdc7/R/datatables.R#L148 However, after working on this PR, this no longer seems to be an important reason, since we can discard the data in the preRenderHook() as well.

Anyway, it does not hurt for server=TRUE to belong to datatable(). Unless you have strong objections, how about just keep it there?

jcheng5:
I guess I'm OK with that.

Are all of these changes going to be difficult to explain in the docs? I guess they only apply to the Shiny page and mostly will involve removing caveats.

yihui:
I don't think they are difficult to explain. We can present the easier syntax primarily (server=TRUE and done) and deemphasize the datatableAjax() thing. I'll merge this PR and update the docs accordingly. Thanks!

PR #90, opened 2015-06-02
Move `server` parameter from datatable to renderDataTable

jcheng5:
I know I said I was OK with us not doing this but I decided it was worth
it after all, for the following benefits:

1. Allows us to default server=TRUE when in a Shiny app
2. Allows server=TRUE to be specified identically whether datatable() or
   data frame is returned in the renderDataTable expr
3. Prevents the nonsensical specification of server=TRUE when using DT
   in a non-Shiny context

The code is actually quite a bit simpler than before I think, though
there is a nontrivial risk of regression.

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -38,9 +38,6 @@
 #'   = FALSE)}, where \code{clear} indicates whether you want the clear buttons
 #'   in the input boxes, and \code{plain} means if you want to use Bootstrap
 #'   form styles or plain text input styles for the text input boxes
-#' @param server whether to use server-side processing; if \code{TRUE}, you must
-#'   provide a server URL so that DataTables can send Ajax requests to retrieve
-#'   data from the server
 #' @param escape whether to escape HTML entities in the table: \code{TRUE} means
 #'   to escape the whole table, and \code{FALSE} means not to escape it;
 #'   alternatively, you can specify numeric column indices or column names to
@@ -65,7 +62,7 @@
 datatable = function(
   data, options = list(), class = 'display', callback = JS('return table;'),
   rownames, colnames, container, caption = NULL, filter = c('none', 'bottom', 'top'),
-  server = FALSE, escape = TRUE, style = 'default',
+  escape = TRUE, style = 'default',
   selection = c('multiple', 'single', 'none'), extensions = list()
 ) {
 
@@ -97,13 +94,6 @@ datatable = function(
   if (length(numc)) options = appendColumnDefs(
     options, list(className = 'dt-right', targets = numc - 1)
   )
-
-  # 1.5Mb is just an arbitrary size from my experiments
-  if (object.size(data) > 1.5e6 && !server && getOption('DT.warn.size', TRUE))
-    warning(
-      'It seems your data is too big for client-side DataTables. You may ',
-      'consider server-side processing: http://rstudio.github.io/DT/server.html'
-    )
 
   # make sure the table is _not_ ordered by default (change the DataTables default)
   if (is.null(options[['order']])) options$order = list()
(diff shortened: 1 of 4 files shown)


R/shiny.R
```diff
@@ -58,11 +64,25 @@ renderDataTable = function(expr, env = parent.frame(), quoted = FALSE, ...) {
...
+      }
     }
 
-    if (is.function(hook <- instance$preRenderHook))
-      instance = hook(instance, currentSession, currentOutputName)
-    instance$preRenderHook = NULL
+    # in the server mode, we should not store the full data in JSON
+    if (server) {
+      origData = instance[['x']][['data']]
+      instance$x$data = NULL
```
wch: Just curious - why the mixing of `instance[['x']][['data']]` and `instance$x$data`?

R/datatables.R
```diff
@@ -225,7 +190,24 @@ datatable = function(
...
+      # TODO: Move this to widget_data override, or preRenderHook
+      if (object.size(data) > 1.5e6 && getOption('DT.warn.size', TRUE))
+        warning(
+          'It seems your data is too big for client-side DataTables. You may ',
+          'consider server-side processing: http://rstudio.github.io/DT/server.html'
+        )
+
+      data = escapeData(data, escape, colnames)
+      data = unname(data)
+      instance$x$data <- data
```
wch: `=`?

R/shiny.R
jcheng5: @yihui recently reminded me that using `instance$x$data` for reads will do partial matching, while `[[` will not. For writes `$` is fine.

I don't know that I'll do `[[` style reads forever, but just today I could not stomach the partial matching... :-\

wch: Oh, that's true. If the items in `instance` are fully under your control, it shouldn't be a problem as long as you do development with `options(warnPartialMatchDollar = TRUE)`. It's probably a good idea in general to also enable `warnPartialMatchArgs` and `warnPartialMatchAttr`.

wch:
I don't have a lot of useful input here, but it looks reasonable to me.

R/shiny.R
jcheng5: Yeah, those are on. I guess it's probably fine, I'll change it.

jcheng5: Actually he seems to use this style (`[[` for read, `$` for write) pretty extensively, so I'll just leave it as-is.

wch: Sounds reasonable.

jcheng5:
Interactive test. (I don't recommend selecting `diamonds` and unchecking `server`...) Some combinations of options will cause warnings, those are intended.

```r
library(shiny)
library(DT)
library(ggplot2)

dataWithHtml <- cbind(diamonds, text = "&nbsp;")

ui <- fluidPage(
  h2("DT test app"),
  sidebarLayout(
    sidebarPanel(
      radioButtons("dataset", "Data set", choices = c(
        "cars (small)" = "cars",
        "diamonds (large)" = "diamonds"
      ), selected = "cars"),
      radioButtons("exprtype", "Expression type", choices = c(
        "raw dataframe" = "df",
        "datatable widget" = "datatable"
      )),
      checkboxInput("server", "server = TRUE?"),
      checkboxInput("escape", "escape = TRUE?")
    ),
    mainPanel(
      DT::dataTableOutput("dt")
    )
  )
)

  df <- reactive({
    switch(input$dataset, cars = cars, diamonds = diamonds)
  })
  
  dfPlusHtml <- reactive({
    cbind(df(), c(HTML = "Foo <strong>bar</strong>"))
  })
  
  dt <- reactive({
    switch(input$exprtype,
      df = dfPlusHtml(),
      datatable = datatable(dfPlusHtml())
    )
  })
  
  observe({
    escape <- input$escape
    server <- input$server
    output$dt <- DT::renderDataTable(dt(), escape = escape, server = server)
  })
}

shinyApp(ui, server)

jcheng5:
Hmmm, crap. I thought `server=TRUE` only made sense in Shiny but @yihui's docs have an example of using it from a non-Shiny context, with a third-party data source: http://rstudio.github.io/DT/server.html

jcheng5:
I'm going to revert until I understand the issue better.

jcheng5:
I've un-reverted. `server=TRUE` is not very useful in non-Shiny contexts, all it does is add `serverSide = TRUE` and `processing = TRUE`.

PR #166, opened 2015-08-13
Defer the table render until the container has non-zero width/height

yihui:
This fixes Tareef's ioslides issue. @jcheng5

### DIFF
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -50,7 +50,16 @@ var transposeArray2D = function(a) {
 HTMLWidgets.widget({
   name: "datatables",
   type: "output",
-  renderValue: function(el, data) {
+  initialize: function(el, width, height) {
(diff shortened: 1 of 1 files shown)


inst/htmlwidgets/datatables.js
```diff
@@ -50,7 +50,16 @@ var transposeArray2D = function(a) {
 HTMLWidgets.widget({
   name: "datatables",
   type: "output",
-  renderValue: function(el, data) {
+  initialize: function(el, width, height) {
+    $(el).text('TO BE RENDERED...');
```
yihui: I can also go without this line, in which case it seems I only need to check `if (el.offsetWidth === 0)` below on line 58, because `height` is still 0 when the resize method is triggered since el is empty.

jcheng5: Can it at least be a nonbreaking space, or an empty div, or something else that wouldn't be so obviously visible?

Other than that, LGTM.

PR #210, opened 2015-11-24
Allow simple comparators for filtering ('=','>=', '<=')

fbreitwieser:
Currently, it is possible to filter for values greater or smaller than a threshold with 'X...' or '...X'. This is not the most intuitive notation, furthermore there is no straightforward way of selecting one value ('X...X'). 

This pull request enables an alternative way of setting the filter in numeric or date-time columns with '>=X', '<=X' and '=X'.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.1.38
+Version: 0.1.39
 Date: 2015-06-08
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


yihui:
Thanks! I guess `X...`, `...X`, and `X...X` are not too bad, especially when you can use the sliders to filter the table.

Anyway, could you squash the 5 commits into one to clean up the history?

```
git reset --soft HEAD~5
# then git add/commit and force push
```

fbreitwieser:
Thank yihui. Please see #277. I opened a new pull request with one clean commit.

PR #243, opened 2016-03-10
fillContainer option (automatically fill parent bounds)

jjallaire:


jjallaire:
@jmcphers We're adding a feature to DT to allow it to fill it's container (just like the RStudio data viewer does). An issue we've run into which I believe you also ran into (and somehow overcame) is that when you set a scrollY on DT and add filtering, the filter popup appears _behind_ the scrollBody of the data table. You can repro this with the following simple Rmd file:

    ---
    title: "DataTables Filtering"
    output: flexdashboard:::flex_dashboard
    ---

    ### DataTable Filtering 

    ```{r, fig.width=7, fig.height=10}
    library(DT)
    datatable(mtcars[,c("mpg", "cyl", "disp")], filter = "top", fillContainer = TRUE)
    ```

Note that to repro this you'll need to install this PR as well as flexdashboard from GitHub:

```r
devtools::install_github("rstudio/flexdashboard")
devtools::install_github("rstudio/DT", ref = "feature/fill-container")
```

Is there a straightforward CSS workaround to this or is the change required more painful?

jmcphers:
It shouldn't be painful; when we built the UX for RStudio we got around this by setting the `z-index: 25` on the filter popup. If that doesn't help, the other major difference between the RStudio configuration and this one is that in RStudio the filter popup is a sibling of the data table rather than being embedded in the scrollBody, which of course makes occluding it much more straightforward (at the expense of losing scroll tracking).

jjallaire:
Thanks for the help Jonathan, here's the full fix: https://github.com/rstudio/DT/commit/36e87224a6e647692fd4342d0abaf8bd0285ef46

jmcphers:
No problem. Looks great!

jjallaire:
@yihui and @jcheng5 Assuming that you both think this implementation and user-interface is reasonable I'd like to get this merged soon so that I can document it on the flexdashbaord website (DT is close to unusable in flexdashboard without this behavior, and I'd rather not ask people to install from a branch). A few notes on the implementation:

1. The reason it's so involved is that while DT has a `scrollY` option for specifying internal scrolling they don't have any built in responsive height capabilities so you basically need to measure the DOM elements in the header and footer every time it's container is resized.

2. There is also some messiness around the timing of when measurement / column width calculation take place (this was all based on documents hacks/workarounds suggested in the DT forums).

3. The new `fillContainer` parameter can be defaulted via an option (`DT.fillContainer`). This is so that responsive containers (e.g. flexdashboard or html slides/storyboard formats) can set it so the user doesn't have to. 

Let me know what you guys think and hopefully we can get this merged soon!

yihui:
Looks good to me. I tested it and it works as expected. Thanks! (@jcheng5 I have merged it and you may post-review it if you like)

jjallaire:
Thanks!

J.J.

PR #277, opened 2016-04-19
Using comparators for filtering table ('=', '>=', '<=')

fbreitwieser:
Currently, it is possible to filter for values with 'X...', '...X' and 'X...X'. 

This pull request enables an alternative way of setting the filter in numeric or date-time columns with '>=X', '<=X' and '=X'.

This is basically the same as pull request #210. I was asked for a combined commit for a nicer history. 

Thanks!

### DIFF
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -332,9 +332,21 @@ HTMLWidgets.widget({
               if ($input.val() === '') filter.val([r1, r2]);
             },
             change: function() {
-              var v = $input.val().replace(/\s/g, '');
-              if (v === '') return;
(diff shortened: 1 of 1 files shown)


yihui:
Thanks for the contribution! As I said in #210, I think the benefit is relatively small. Another problem is when you move the slider, these expressions will still be replaced by `x1 ... x2` in the text box.

fbreitwieser:
That's true, it's a small thing, and not entirely consistent. But it's usage feels a bit more natural to me.

No problem if you decide not to merge the commit, though! It's a request after all ;)

yihui:
Yes, I definitely agree it is more natural.

Issue #276, opened 2016-04-19
formatPercentage: Don't fix the digits


fbreitwieser:
I have data I would like to have displayed as percent, however without a fixed number of digits. (I do the rounding beforehand using `signif`, which is more appropriate for this data.)

Can you change the functionality such that when `digits=NULL`, the number is not set as fixed? Or, alternatively, provide a convenience function to just modify a column value with a `before` and `after`?

The code where the number of digits is set is in `datatables.js` in `DTWidget.formatPercentage`: 
`$(thiz.api().cell(row, col).node()).html((d * 100).toFixed(digits) + '%');`

fbreitwieser:
I implemented a function `formatString(table, columns, string_before, string_after)`. See https://github.com/fbreitwieser/DT/commit/4aa624e24088c598d7bae0a05b02818aecb1616a . I'm happy to send a pull request.

Usage:
`dt %>% formatString(1, string_before='In percent: ', string_after='%')`

PR #279, opened 2016-04-20
Added function formatString to prepend and append to column values

fbreitwieser:
Currently there's no convenience function to just append or prepend values to the column values.

This commit implements a function `formatString` that takes `string_before` and `string_after` values.

`formatString(table, columns, string_before, string_after).`

Usage:
 dt %>% formatString(1, string_before='This value: ', string_after='%')

Initially mentioned in #276


yihui:
Thanks! I merged the PR but renamed the arguments to `prefix` and `suffix`, respectively.

PR #314, opened 2016-06-28
add selectPage

carlganz:
Apropos of this stackoverflow question: http://stackoverflow.com/questions/37891616/how-do-i-switch-the-page-displayed-in-a-shiny-datatable

I know it isn't a super useful function, but I thought I'd share.

Regards


yihui:
I think it is useful. Thanks!

PR #434, opened 2017-06-22
Make test snapshots consistent between runs

wch:
This uses the new `shiny::snapshotPreprocess` function to remove the AJAX URL from the test snapshot. This is for rstudio/shinytest#107.

Note that because `shiny::snapshotPreprocess` is a new function, DT should be released only after the next version of Shiny goes to CRAN; otherwise it will raise a warning with R CMD check on the CRAN build servers.


yihui:
Thanks Winston!

Issue #28, opened 2015-03-22
Is there an Editor extension available for the DT package


mtseman:
I love this package. I combined several extensions and it really helps my app look great and function.

I thought KeyTable would allow me to edit, but it just navigates and highlights.

Is there an editor extension available or coming soon.

Thanks!
Mike

yihui:
I guess it is probably not coming soon, but thanks for filing the request anyway :)

mtseman:
Hi Yihui,

Just a question. If I bought the Editor from DT, would it be easy to make
it work in a shinyapp? I recall your webinar had some instructions for
creating widgets outside

I looked at their site I need client and server side code(?). As I
understand it, I saw there was downloads for javascript (client side) and
.NET or PHP (server side) (?)

What would be a good resource or recommended material would you suggest I
look into to understanding more how this works.

Thanks,
Mike

yihui:
@mtseman Sorry, I have not really played with the Editor yet, so I do not know how to answer you at the moment. I _imagine_ we are not going to need .NET or PHP, and editing can be done through R if you use server-side processing. It should not be too difficult to implement a prototype, but I'm not sure what other features that users might want besides editing the values in cells.

danielkrizian:
It would be great to have Editor extension available. One use case is user input of tabular data, see one mockup here: https://github.com/danielkrizian/handsonimport, which I implemented with [Handsontable](http://handsontable.com/).

jpmarindiaz:
+1 
@danielkrizian there is also this html widget: https://github.com/jrowen/rhandsontable
It does the trick for now, but it would be pretty cool to have the edit functionality in DT!

yihui:
This is a built-in feature of Handsontable but not DataTables (I did not purchase the Editor extension or try it). I will have to do every bit by myself, and that is what has been holding me here. I just outlined some ideas in #64.

shrektan:
Hey, YiHui, I like your datatable package. I will vote for the idea to add an editor extension, since after some exploration, I didn't see any easy-to-use way to edit the table, except the `editor`.

Thanks for what you done, for all these R packages that riches the life~!

ihaddad-md:
Strong +1 :)

maxmoro:
:+1: me too

yihui:
I have been experimenting with this recently, and I have a working prototype now. You can double click on a cell to edit its value:

```r
devtools::install_github('rstudio/DT@feature/editor')

library(shiny)
library(DT)
[...]

PR #475, opened 2018-01-07
support shift+click to select mutiple rows

### LINKED ISSUE #305: Feature request: Selecting multiple rows using shift+click
Thank you for the wonderful package.

It would be very useful to be able to select multiple rows using shift+click instead of having to click each
row individually.

I see that this has been asked for [previously](https://groups.google.com/forum/#!topic/shiny-discuss/PdhVoX0FJlU).

shrektan:
closes #305

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 - Added the `RowGroup` extension (thanks, @ecortens, #466 #442).
 
+- Support using `shift + click` to select multiple rows (@shrektan, #475 #305).
+
 ## MAJOR CHANGES
 
 - For client-side tables in Shiny, the indices of selected rows (`input$tableId_rows_selected`) will include the rows that are not visible in the current view after filtering is applied. If you want to exclude the indices of such rows, you can take the intersection of `input$tableId_rows_selected` and `input$tableId_rows_all` (the latter is the indices of all rows after filtering is applied).
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 - Added the `RowGroup` extension (thanks, @ecortens, #466 #442).
 
-- Support using `shift + click` to select multiple rows (@shrektan, #475 #305).
+- Support using `Shift + Click` to select multiple rows (thanks, @shrektan, #475 #305).
 
 ## MAJOR CHANGES
 
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -741,7 +741,38 @@ HTMLWidgets.widget({
         selected1 = $.makeArray(selected.rows);
         selected2 = $.makeArray(selected.cols);
       }
+
+      // After users reorder the rows or filter the table, we cannot use the table index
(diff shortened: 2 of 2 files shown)


shrektan:
Here's my testing code.

```r
library(shiny)

server <- function(input, output) {
  output$data <- DT::renderDT(
    DT::datatable(iris), server = F
  )
  output$data_sel <- renderText(input$data_rows_selected)
}
ui <- fluidPage(
  tagList(
    DT::DTOutput("data"),
    textOutput("data_sel")
  )
)

shinyApp(ui, server, options = list(port = 7995))

```

yihui:
I have fixed the text selection issue. The other thing is that this probably does not work in the server-processing mode (which is always tricky in DT)...

shrektan:
Well, just confirm that there's indeed a small issue (seem like small), which is the selection range will be different between the server-processing mode and the client side mode.

1. select first row in page 1, let's assume it's row no 1;
1. press <kbd>shift</kbd>
1. select another row in another page, let's assume it's row no 20, and every page displays 5 row

In the client side mode, the row selection will be 1 to 20, while in the server-processing mode, it's 1, 16 to 20...

yihui:
You need to use the vector `DT_rows_current` to convert the local row index to the index on the server side. You can see how I did it in `changeInput('rows_selected')` and `changeInput('row_last_clicked')`. Let me know if you are able to fix it. Thanks!

shrektan:
I guess I should be able to fix it based on your current code, some ideas on server-processing mode:

+ `lastClickedRow` and `currentClickedRow` (i.e., `thisRow.index()`) should be converted to the index on server side;
+ `shiftSelRows()` should return the indexes on server side as well. Since on client side, the browser itself doesn't know what the real underlying data is, we can only convert the indexes relying on `DT_rows_current`.
+ Select the rows based on server side indexes. Here's the tricky part, how can I be able to select rows that hasn't been loaded to the browsers yet? By looking at your implementation for [`selection = list(mode = "multiple", selected = c(1, 3, 8)`](https://github.com/shrektan/DT/blob/421c18da7503942a486f9f7dfbd84506dcff62e6/inst/htmlwidgets/datatables.js#L826-L835)  I think it can be done by changing the value of `selected1`.

Will try to fix it tonight or tomorrow~

yihui:
Sounds great!

yihui:
我的 0.3 版本计划的事项现在还剩下两项：https://github.com/rstudio/DT/issues?q=is%3Aopen+is%3Aissue+milestone%3Av0.3 我先做 #28 去了，你这个事项需要比较细心的考虑，所以你慢慢来不着急，下周之前能捣鼓出来最好，捣鼓不出来也没事。多谢！

shrektan:
恩恩，白天上班没什么时间 :man_office_worker: 

昨天稍微鼓捣了下，要是没有幺蛾子的话，感觉应该可以弄出来。因为没有测试嘛，代码有无其他坏的影响，姑且只能由你来把关啦，哈哈哈。

yihui:
嗯，我也只能用意念力在脑子里“测试”。

PR #478, opened 2018-01-13
fix #476 

### LINKED ISSUE #476: dt-right class is added to numeric column headers
I am fairly certain this problem was recently introduced. It just recently appeared in a complex application of mine, but I have figured out what it boils down to.

MRE:

```r
datatable(head(iris), 
options = list(columnDefs = list(list(
targets = 0, 
className = "dt-center"))), 
rownames = FALSE)
```
It's hard to notice on smaller screen, but header for first column has dt-right class so dt-center isn't fully respected. Here is an image:

dt-bug

I haven't paid enough attention to recent changes to know why this is happening. Would greatly appreciate if column title respected className option. 

Kind Regards

shrektan:
Setting numeric columns alignment to `dt-right` should respect users' option.

This issue arises from the updating version of `DataTable` https://github.com/rstudio/DT/commit/9235453ef3825525eb4dd861ce4e4de620bad633

It seems like in the old version, `DataTable` only uses the first `className` for multiple defined columns. While in the new version, it will attach multiple `className` to the same column in order and causes the issue.

closes #476

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -125,9 +125,15 @@ datatable = function(
   }
 
   # align numeric columns to the right
-  if (length(numc)) options = appendColumnDefs(
-    options, list(className = 'dt-right', targets = numc - 1)
-  )
+  if (length(numc)) {
+    # if the `className` of the column has already been defined by the user,
+    # we should not touch it.
+    defined_cols <- classNameDefinedColumns(options)
+    undefined_numc <- (numc - 1)[!(numc - 1) %in% defined_cols]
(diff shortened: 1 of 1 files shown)


shrektan:
```r
datatable(
  head(iris),
  options = 
    list(columnDefs = list(
      list(targets = 0, className = "dt-center"),
      list(targets = 1, className = "dt-left"),
      list(targets = 2, className = "dt-right")
    )),
  rownames = FALSE
)
```

# output before



# output now

PR #480, opened 2018-01-17
Make it possible to edit values in table

yihui:
Closes #28

## Test

```r
devtools::install_github('rstudio/DT@feature/editor')
```

Double click in a table cell to edit its value.

## (1) client-side processing (static HTML application)

```r
DT::datatable(iris)
```

## (2) client-side processing in Shiny

```r
library(shiny)
library(DT)
shinyApp(
  ui = fluidPage(
    DT::dataTableOutput('x2')
  ),
  server = function(input, output, session) {
    x = iris
    x$Date = Sys.time() + seq_len(nrow(x))
    output$x2 = DT::renderDataTable(x, selection = 'none', server = FALSE)
  }
)
```

## (3) server-side processing

```r
library(shiny)
library(DT)
shinyApp(
  ui = fluidPage(
    DT::dataTableOutput('x1')
  ),
  server = function(input, output, session) {
    x = iris
    x$Date = Sys.time() + seq_len(nrow(x))
    output$x1 = DT::renderDataTable(x, selection = 'none')
    
    proxy = dataTableProxy('x1')
    
    observeEvent(input$x1_cell_edit, {
      info = input$x1_cell_edit
      str(info)
      i = info$row
      j = info$col
      v = info$value
      x[i, j] <<- DT::coerceValue(v, x[i, j])
      replaceData(proxy, x, resetPaging = FALSE)  # important
    })
  }
)
```

## (4) server-side processing (without row names)

```r
library(shiny)
library(DT)
shinyApp(
  ui = fluidPage(
    DT::dataTableOutput('x1')
  ),
  server = function(input, output, session) {
    x = iris
    x$Date = Sys.time() + seq_len(nrow(x))
    output$x1 = DT::renderDataTable(x, selection = 'none', rownames = FALSE)
    
    proxy = dataTableProxy('x1')
    
    observeEvent(input$x1_cell_edit, {
      info = input$x1_cell_edit
      str(info)
      i = info$row
      j = info$col + 1  # column index offset by 1
      v = info$value
      x[i, j] <<- DT::coerceValue(v, x[i, j])
      replaceData(proxy, x, resetPaging = FALSE, rownames = FALSE)
    })
  }
)
```

PR #484, opened 2018-01-18
add editable parameter

### LINKED ISSUE #483: Please don't make tables editable by default
Hello @yihui,

The new editor feature is great. However, the current implementation makes all tables editable, and I don't see a parameter that lets you opt out.

I will make a PR that makes the editor optional. I think it would be a good idea to make the table uneditable by default. There are many datatables in production Shiny applications that were built with the expectation that the tables would not be editable. 

Kind Regards

carlganz:
When I run roxygen all the rd files get rewritten so I didn't update docs

Feel free to disregard this PR if you have a different implementation.

Fixes #483

### DIFF
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -668,28 +668,30 @@ HTMLWidgets.widget({
     // run the callback function on the table instance
     if (typeof data.callback === 'function') data.callback(table);
 
+    if (data.editable) {
     // double click to edit the cell
     table.on('dblclick.dt', 'tbody td', function() {
-      var $input = $('<input type="text">');
-      var $this = $(this), value = table.cell(this).data(), html = $this.html();
-      $input.val(value);
-      $this.empty().append($input);
-      $input.css('width', '100%').focus().on('change', function() {
-        var valueNew = $input.val();
-        if (valueNew != value) {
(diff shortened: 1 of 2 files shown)


inst/htmlwidgets/datatables.js
```diff
@@ -668,28 +668,30 @@ HTMLWidgets.widget({
     // run the callback function on the table instance
     if (typeof data.callback === 'function') data.callback(table);
 
+    if (data.editable) {
```
yihui: This PR looks good to me, but I hope to avoid the changes from indentation. You don't need the curly braces since it is a single expression, e.g.

```js
if (data.editable) table.on('dblclick.dt', 'tbody td', function() {
  ...
```

carlganz:
Awesome thanks!

danielcanueto:
Would it be possible to make that edition is not only usable for visual purposes but that actual values are also changed? reactiveValues variables visualized through DT can be edited on the table but edits are not saved to the variables. This was possible with previous feature/editor branch, but it seems this brach is not active anymore.

yihui:
@danielcanueto Did you read the examples in #480?

danielcanueto:
Sorry, I hadn't seen this pull. Thanks!

PR #498, opened 2018-02-09
Respond to plotly's persistent selection mode

cpsievert:
This adds the ability to perform persistent selection with plotly (similar to how https://github.com/rstudio/leaflet/pull/346 does for leaflet).

It would be ideal to have rows somehow reflect [plotly's selection color](https://github.com/rstudio/leaflet/pull/346/files#diff-8e0e6e242b99825df0596336e5ad4d7aR876), let me know if have any ideas how best to go about that (e.g. color the text or highlight the row?)

dt

### DIFF
--- a/inst/examples/DT-crosstalk/plotly-persist.R
+++ b/inst/examples/DT-crosstalk/plotly-persist.R
@@ -1,0 +1,15 @@
+library(plotly)
+library(crosstalk)
+library(DT)
+
+m <- SharedData$new(mtcars)
+bscols(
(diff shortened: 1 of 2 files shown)


cpsievert:
Cool, thanks! BTW, adding color can be useful for making comparisons, so I would say we should add it at some point, but I might not have the time for it in the near future, want me to raise an issue?

dt2

yihui:
I think it is a very cool idea, although I definitely don't have time, either :)

PR #503, opened 2018-02-13
accumulate matches (rather than matches and keys); fixes #502

### LINKED ISSUE #502: Persistent selection doesn't highlight correct keys
Looks like #498 was `concat()` objects which lead to the incorrect keys being highlighted (notice how Porsche should be included in the data table in that gif)

PR coming up....

cpsievert:
dt-fix


yihui:
Thanks!

PR #522, opened 2018-04-05
Improve the format of the filter of `Date` / `Datetime` columns

### LINKED ISSUE #241: Timezone in Filter
Hi there,

I am new to DT and I am trying to figure out how to set time zone in the filter for a datatable in a Shiny app. The table has a column of time stamps. I'd like it to be in US/Central and it seems to work well in the table. However the filter is shown in UTC. 

Here is an example:

``` R
library(DT)
library(shiny)
Sys.setenv(TZ = "US/Central")

server <- function(input, output) {
  output$mytable = DT::renderDataTable(
    data.frame(time = as.POSIXct("2016-02-29 13:30:00",tz="US/Central")+c(1:5), value=c(1:5))
  , filter = 'top')
}

ui <- fluidPage(
  DT::dataTableOutput('mytable')
)

shinyApp(ui = ui, server = server)
```

image

shrektan:
Closes #241 

Improve the format of the filter of `Date` / `Datetime` columns. It will display the same format and timezone as the column content.

~~At the time of writing, the PR is unfinished because it doesn't work for server-side processing mode.~~ DONE, it now works properly with server-side processing mode.

# TODO

- [x] ensure the filter returns correct contents in server-side processing mode.

# Example 

```r
library(DT)
library(shiny)

df <- data.frame(
  time = lubridate::ymd_hm(201801011000) + lubridate::dminutes(0:120),
  time2 = lubridate::ymd_hm(201801011000) + lubridate::dminutes(0:120),
  date = lubridate::ymd(20180101) + lubridate::ddays(0:120),
  date2 = lubridate::ymd(20180101) + lubridate::ddays(0:120)
)

ui <- fluidPage(
  h3("server = TRUE"),
  DT::DTOutput("tbl1"),
  hr(),
  h3("server = FALSE"),
  DT::DTOutput("tbl2"),
  hr(),
  h3("without `formatDate()`"),
  DT::DTOutput("tbl0")
)

  output$tbl0 <- DT::renderDT({
    datatable(df, filter = "top")
  })
  output$tbl1 <- DT::renderDT({
    datatable(df, filter = "top") %>% 
      formatDate(c("time"), method = 'toLocaleString', params = list("en-US", list(timeZone = "Asia/Shanghai", hour12 = FALSE))) %>%
      formatDate(c("time2"), method = 'toLocaleString', params = list("se", list(timeZone = "UTC", hour12 = FALSE))) %>%
      formatDate(c("date"), method = 'toLocaleDateString', params = list("en-US")) %>%
      formatDate(c("date2"), method = 'toLocaleDateString')
  }, server = TRUE)
  output$tbl2 <- DT::renderDT({
    datatable(df, filter = "top") %>% 
      formatDate(c("time"), method = 'toLocaleString', params = list("en-US", list(timeZone = "Asia/Shanghai", hour12 = FALSE))) %>%
      formatDate(c("time2"), method = 'toLocaleString', params = list("se", list(timeZone = "UTC", hour12 = FALSE))) %>%
      formatDate(c("date"), method = 'toLocaleDateString', params = list("en-US")) %>%
      formatDate(c("date2"), method = 'toLocaleDateString')
  }, server = FALSE)
}

shinyApp(ui, server)

```

### DIFF
--- a/R/format.R
+++ b/R/format.R
@@ -111,7 +111,7 @@ formatDate = function(table, columns, method = 'toDateString', params = NULL) {
     if (is.null(params)) params = list()
     cols = as.character(name2int(columns, colnames, rownames))
     x$filterDateFmt = as.list(x$filterDateFmt)
-    for (col in cols) x$options$filterDateFmt[[col]] = list(
+    for (col in cols) x$filterDateFmt[[col]] = list(
       method = method, params = toJSON(params)
     )
     table$x = x
--- a/R/format.R
+++ b/R/format.R
@@ -110,7 +110,7 @@ formatDate = function(table, columns, method = 'toDateString', params = NULL) {
     rownames = base::attr(x, 'rownames', exact = TRUE)
     if (is.null(params)) params = list()
     cols = as.character(name2int(columns, colnames, rownames))
-    if (is.null(x$options$filterDateFmt)) x$options$filterDateFmt = list()
+    x$filterDateFmt = as.list(x$filterDateFmt)
     for (col in cols) x$options$filterDateFmt[[col]] = list(
       method = method, params = toJSON(params)
     )
--- a/R/format.R
+++ b/R/format.R
@@ -109,7 +109,7 @@ formatDate = function(table, columns, method = 'toDateString', params = NULL) {
     colnames = base::attr(x, 'colnames', exact = TRUE)
     rownames = base::attr(x, 'rownames', exact = TRUE)
     if (is.null(params)) params = list()
-    cols = as.character(name2int(columns, colnames, rownames))
+    cols = sprintf("%d", name2int(columns, colnames, rownames))
     x$filterDateFmt = as.list(x$filterDateFmt)
     for (col in cols) x$filterDateFmt[[col]] = list(
       method = method, params = toJSON(params)
--- a/R/format.R
+++ b/R/format.R
@@ -103,6 +103,21 @@ formatSignif = function(
 #'   \code{params = list('ko-KR', list(year = 'numeric', month = 'long', day =
 #'   'numeric'))}
 formatDate = function(table, columns, method = 'toDateString', params = NULL) {
+  x = table$x
+  if (x$filter != 'none') {
+    if (inherits(columns, 'formula')) columns = all.vars(columns)
+    colnames = base::attr(x, 'colnames', exact = TRUE)
+    rownames = base::attr(x, 'rownames', exact = TRUE)
+    if (is.null(params)) params = list()
+    cols = as.character(name2int(columns, colnames, rownames))
+    if (is.null(x$options$filterDateFmt)) x$options$filterDateFmt = list()
+    for (col in cols) x$options$filterDateFmt[[col]] = list(
+      method = method, params = toJSON(params)
+    )
+    table$x = x
+  }
+  # the code above is used to ensure the date(time) filter displays the same format or
+  # timezone as the column value
   formatColumns(table, columns, tplDate, method, params)
 }
 
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -563,7 +563,10 @@ HTMLWidgets.widget({
             }
             r1  = t1; r2 = t2;
           })();
-          $span1.text(formatDate(r1, false)); $span2.text(formatDate(r2, false));
+          var updateSliderText = function(v1, v2) {
+            $span1.text(formatDate(v1, false)); $span2.text(formatDate(v2, false));
+          };
+          updateSliderText(r1, r2);
           var updateSlider = function(e) {
             var val = filter.val();
             // turn off filter if in full range
@@ -575,7 +578,7 @@ HTMLWidgets.widget({
             } else {
               $input.attr('title', '').val('');
             }
-            $span1.text(formatDate(val[0], false)); $span2.text(formatDate(val[1], false));
+            updateSliderText(val[0], val[1]);
             if (e.type === 'slide') return;  // no searching when sliding only
             if (server) {
               table.column(i).search($td.data('filter') ? ival : '').draw();
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -563,7 +563,7 @@ HTMLWidgets.widget({
             }
             r1  = t1; r2 = t2;
           })();
-          $span1.text(formatDate(r1)); $span2.text(formatDate(r2));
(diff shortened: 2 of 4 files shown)


shrektan:
I didn't notice it before... Yes, it's not always possible to revert it back so there's no perfect solution. 

I think your suggestion is acceptable but it will be better with one little improvement, which is to display the time with the same timezone as it's been set in the column value.

shrektan:
It seems like quite difficult to deal with timezone issues in javascript... I will temporarily give up the idea of displaying the same timezone... 😢 

Now the textbox displays UTC value so it means we can change the text to update the sliders now.

inst/htmlwidgets/datatables.js
```diff
@@ -520,11 +520,14 @@ HTMLWidgets.widget({
               filter.val(v);
             }
           });
-          var formatDate = function(d) {
+          var formatDate = function(d, isoFmt = false) {
```
yihui: Default parameter values are not supported in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters. The conservative approach is to test if `isoFmt` is `undefined`.

R/format.R
```diff
@@ -103,6 +103,21 @@ formatSignif = function(
...
 #'   'numeric'))}
 formatDate = function(table, columns, method = 'toDateString', params = NULL) {
+  x = table$x
+  if (x$filter != 'none') {
+    if (inherits(columns, 'formula')) columns = all.vars(columns)
+    colnames = base::attr(x, 'colnames', exact = TRUE)
+    rownames = base::attr(x, 'rownames', exact = TRUE)
+    if (is.null(params)) params = list()
+    cols = as.character(name2int(columns, colnames, rownames))
+    if (is.null(x$options$filterDateFmt)) x$options$filterDateFmt = list()
```
yihui: I'd store the info in `x$filterDateFmt` instead of `x$options$filterDateFmt`, because `x$options` is for initializing a DataTable via `.DataTable()`, and `filterDateFmt` is not one of the supported options (which is why the current PR can work):
https://github.com/rstudio/DT/blob/8ba54abd4e4b1215908aae83981f4b51d3be1984/inst/htmlwidgets/datatables.js#L269

Accordingly, on the JS side, use `data.filterDateFmt` instead of `options.filterDateFmt`.

inst/htmlwidgets/datatables.js
```diff
@@ -566,16 +569,17 @@ HTMLWidgets.widget({
...
+              ival = formatDate(val[0], true) + ' ... ' + formatDate(val[1], true);
               $input.attr('title', ival).val(ival).trigger('input');
             } else {
               $input.attr('title', '').val('');
             }
-            $span1.text(v1); $span2.text(v2);
+            $span1.text(formatDate(val[0])); $span2.text(formatDate(val[1]));
             if (e.type === 'slide') return;  // no searching when sliding only
             if (server) {
+              ival = formatDate(val[0], true) + ' ... ' + formatDate(val[1], true);
```
yihui: There is some redundancy in these changes from line 572 to 582. I'll revise the PR by myself to show you.

R/format.R
```diff
@@ -103,6 +103,21 @@ formatSignif = function(
 #'   \code{params = list('ko-KR', list(year = 'numeric', month = 'long', day =
 #'   'numeric'))}
 formatDate = function(table, columns, method = 'toDateString', params = NULL) {
+  x = table$x
+  if (x$filter != 'none') {
+    if (inherits(columns, 'formula')) columns = all.vars(columns)
+    colnames = base::attr(x, 'colnames', exact = TRUE)
+    rownames = base::attr(x, 'rownames', exact = TRUE)
+    if (is.null(params)) params = list()
+    cols = as.character(name2int(columns, colnames, rownames))
```
yihui: The `as.character()` trick is very subtle here. It relies on the fact of how JS indexes objects: for an object `x` and an integer index `i`, `x[i]` indexes the element with the character name `i`, instead of the `i`-th element. That is how the line 528 works: https://github.com/rstudio/DT/pull/522/files#diff-2c60c8a789627a2716b5b4271c2b398bR528

This trick can be fragile when the number of columns is larger than 100000, because `as.character(100000)` gives you `"1e+5"`. But I imagine it should be extremely rare that users have more than 100 thousand columns in a table.

shrektan: Ha, I didn't notice this. Using `sprintf("%d", 100000)` should be robust enough.

inst/htmlwidgets/datatables.js
shrektan: Thanks for sharing this.

R/format.R
```diff
@@ -110,7 +110,7 @@ formatDate = function(table, columns, method = 'toDateString', params = NULL) {
     rownames = base::attr(x, 'rownames', exact = TRUE)
     if (is.null(params)) params = list()
     cols = as.character(name2int(columns, colnames, rownames))
-    if (is.null(x$options$filterDateFmt)) x$options$filterDateFmt = list()
```
shrektan: Excellent :+1:

inst/htmlwidgets/datatables.js
```diff
@@ -520,14 +520,13 @@ HTMLWidgets.widget({
...
+          var formatDate = function(d, isoFmt) {
             d = scaleBack(d, scale);
             if (type === 'number') return d;
             if (type === 'integer') return parseInt(d);
             var x = new Date(+d);
-            var fmt = undefined;
-            if ('filterDateFmt' in options) fmt = options.filterDateFmt[i];
-            if (!isoFmt && fmt !== undefined) return x[fmt.method].apply(x, fmt.params);
+            var fmt = ('filterDateFmt' in data) ? data.filterDateFmt[i] : undefined;
+            if (fmt !== undefined && isoFmt === false) return x[fmt.method].apply(x, fmt.params);
```
shrektan: shorter and cleaner.

shrektan:
It looks great. Thanks for sharing your ideas by providing the detailed comments. It's very useful.

R/format.R
```diff
@@ -109,7 +109,7 @@ formatDate = function(table, columns, method = 'toDateString', params = NULL) {
     colnames = base::attr(x, 'colnames', exact = TRUE)
     rownames = base::attr(x, 'rownames', exact = TRUE)
     if (is.null(params)) params = list()
-    cols = as.character(name2int(columns, colnames, rownames))
+    cols = sprintf("%d", name2int(columns, colnames, rownames))
```
yihui: Yes, this is more robust, but will still have problems if a column index is too big (bigger than 2^31). Again, I think it is going to be really rare, so that is fine to me.

PR #532, opened 2018-04-22
the filter & server mode now play nicely with the colReorder extension

### LINKED ISSUE #531: Filter + colReorder don't work together
When columns are reordered (whether by the new colReorder function or by a user dragging them around), I think the filters don't apply to the right columns.

The behavior depends on whether server is set, but neither case works correctly. When running the following code, and then dragging the "Species" column to the first position, if I perform a filter and server is set to FALSE, there are no values displayed. If server is set to TRUE, the filter works, but then the data gets filled in ignoring the new order (but the filter works).

Is this fixable in DT? Or is it inherited from the javascript upstream?

```r
library(shiny)
library(DT)

ui <- basicPage(
  DT::dataTableOutput('foo')
)

  output$foo = DT::renderDataTable(
    iris, server = FALSE, # Or server set to TRUE
    filter = "top", 
    caption = 'colReorder + filter can have trouble',
    extensions = 'ColReorder',
    options = list(colReorder = TRUE)
  )
})

runApp(shinyApp(ui, server))
```

## server =FALSE
screen shot 2018-04-17 at 6 39 49 pm
screen shot 2018-04-17 at 6 40 10 pm
screen shot 2018-04-17 at 6 40 16 pm

## server = TRUE
screen shot 2018-04-17 at 6 45 39 pm
screen shot 2018-04-17 at 6 45 45 pm
screen shot 2018-04-17 at 6 45 51 pm

shrektan:
Closes #531 

### Description

Before this PR, if the user enables the `colReorder` extension, the filter will return wrong results in the client mode. In the server processing mode, the table will be reverted to the original order whenever it gets re-drawed (i.e., triggered by the filter, search or the paginate button)

### The causes

- __Client Mode__: The row data returned by the `datatable` is reordered but the filter functions itself is bind to the original order. So the wrong data gets filtered;
- __Server Mode__: Since the data processing is done in R, the row data is not reordered. However, because R doesn't know the columns have been reordered, it returns the data with column unchanged.

### After this PR

It should work for both cases.

### The example

```r
library(shiny)
library(DT)

ui <- fluidPage(
  h1("server - FALSE"),
  fluidRow(
    column(
      6,
      DT::dataTableOutput('tbl_with_colreorder')
    ),
    column(
      6,
      DT::dataTableOutput('tbl_wo_colreorder')
    )
  ),
  h1("server - TRUE"),
  fluidRow(
    column(
      6,
      DT::dataTableOutput('tbl_with_colreorder_s')
    ),
    column(
      6,
      DT::dataTableOutput('tbl_wo_colreorder_s')
    )
  )
)

  output$tbl_with_colreorder = DT::renderDataTable(
    iris, server = FALSE,
    filter = "top", 
    caption = 'with ColReorder',
    extensions = 'ColReorder',
    options = list(colReorder = TRUE)
  )
  output$tbl_wo_colreorder = DT::renderDataTable(
    iris, server = FALSE,
    filter = "top", 
    caption = 'w/o ColReorder'
  )
  output$tbl_with_colreorder_s = DT::renderDataTable(
    iris, server = TRUE,
    filter = "top", 
    caption = 'with ColReorder',
    extensions = 'ColReorder',
    options = list(colReorder = TRUE)
  )
  output$tbl_wo_colreorder_s = DT::renderDataTable(
    iris, server = TRUE,
    filter = "top", 
    caption = 'w/o ColReorder'
  )
}

runApp(shinyApp(ui, server))

```


inst/htmlwidgets/datatables.js
```diff
@@ -257,15 +257,28 @@ HTMLWidgets.widget({
...
     var thiz = this;
     if (instance.fillContainer) $table.on('init.dt', function(e) {
       thiz.fillAvailableHeight(el, $(el).innerHeight());
     });
-
+    // If the page contains serveral datatables and one of which enables colReorder,
+    // the table.colReorder.order() function will exist but throws error when called.
+    // So it seems like the only way to know if colReorder is enabled or not is to
+    // check the options.
+    var colReorderEnabled = function() { return "colReorder" in options; };
```
yihui: Does it have to be a function? I mean, does this work?

```js
var colReorderEnabled = "colReorder" in options;
```

inst/htmlwidgets/datatables.js
```diff
@@ -257,15 +257,28 @@ HTMLWidgets.widget({
...
-        return json.data;
+        var data = json.data;
+        if (!colReorderEnabled()) return data;
+        var table = $table.DataTable(), order = table.colReorder.order(), flag = true, i, j, row;
+        for (i = 0; i < order.length; ++i) if (order[i] !== i) flag = false;
+        if (flag) return data;
+        for (i = 0; i < data.length; ++i) {
+          row = data[i].slice();
+          for (j = 0; j < order.length; ++j) data[i][j] = row[order[j]];
+        }
```
yihui: I wonder if there is a way to reorder the data on the server side. If there is, the implementation could be much easier (`data[order]` in R instead of a nested loop in JS). Of course, it depends on whether we have the order information on the server side.

inst/htmlwidgets/datatables.js
shrektan: Actually, I used the non-function version at first and worked fine... 

However, I overthinked that the `options` may get changed somewhere else after the table being initialized... It's OK to change it.

inst/htmlwidgets/datatables.js
yihui: In that case, let's just forget it. I'm not too concerned about the performance of the nested loops here since `data` should typically be small enough.

yihui:
Just a reminder to myself -- I need to update the docs at https://rstudio.github.io/DT/extensions.html

inst/htmlwidgets/datatables.js
shrektan: I had the concern too... For the deep copy and nested loops... but as you said, it should be fine in the most cases.

PR #558, opened 2018-06-15
add the param default for `styleEqual()`

### LINKED ISSUE #546: Support a default style in styleEqual
For all values that are not specified, use the default style. For example, set background color to green when 0 and red for all other values.

shrektan:
Closes #546

### DESCRIPTION

Before this PR, you can't set the default css value for a column, even if you try to do it like below (I don't why...).

``` r
library(DT)
df <- data.frame(a = 0:9)
tmp <- datatable(df) %>%
  DT::formatStyle("a", backgroundColor = styleEqual(0, "green")) %>%
  DT::formatStyle("a", backgroundColor = "red")
cat(tmp$x$options$rowCallback)
#> function(row, data) {
#> var value=data[1]; $(this.api().cell(row, 1).node()).css({'background-color':'red'});
#> var value=data[1]; $(this.api().cell(row, 1).node()).css({'background-color':value == 0.000000 ? 'green' : ''});
#> }
tmp
```

Anyway, it makes sense to me to be able to set the default CSS value.

### Example 

```r
library(DT)
df <- data.frame(a = 0:9)
datatable(df) %>%
  DT::formatStyle("a", backgroundColor = styleEqual(0, "green", default = "red"))
datatable(df) %>%
  DT::formatStyle("a", backgroundColor = styleEqual(0, "green"))
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.4.12
+Version: 0.4.14
 Authors@R: c(
     person("Yihui", "Xie", email = "xie@yihui.name", role = c("aut", "cre")),
     person("Joe", "Cheng", role = "ctb"),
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.4.13
+Version: 0.4.14
```
yihui: It is fine if you don't bump the version number in each PR. I can do it by myself.

PR #584, opened 2018-08-23
Fix the bug that using DT as a htmlwidget inside UI throws an error

### LINKED ISSUE #583: Using DT as a htmlwidget inside UI throws an error
Here is a minimal reproducible example.

```r
ui <- fluidPage(DT::datatable(iris))
server <- function(input, output, session){}
shinyApp(ui = ui, server = server)
```

This displays a table only with the headers and the following error in the JS console.

image

shrektan:
closes #583

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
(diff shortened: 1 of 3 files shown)


shrektan:
I can't see there's any danger. So I will boldly merge it by myself.

shrektan:
:D

PR #606, opened 2018-10-12
Fix #603: integrate the plugin "ellipsis", get warning if plugin is not supported

### LINKED ISSUE #603: Please integrate the plugin "ellipsis"
Link:
https://datatables.net/plug-ins/dataRender/ellipsis

Also, I have a question: I did not read the fine-print and just tried `DT::datatable(..., plugin = "ellipsis")`, which gave me an error which I found hard to interpret (when using `knitr`; in Rstudio the viewer simply remained empty): "Path for html_dependency not found"

Would it be possible to have a more meaningful error message whenever a plugin is not integrated? Also, it would be nice to explain this in the documentation of the `plugins` option of `DT::datatable`.

jarauh:


djbusstop:
Hello 2 years later! We are using the code example above to implement the ellipsis plugin. Glad we found it! Would probably be good to have in the docs tho, because it doesn't show in the docs that one can use `JS(...)` in the render arg. Just an idea.

PR #627, opened 2018-12-17
DT-626 - fixed issue with non-working reloadData/replaceData 

### LINKED ISSUE #626: replaceData in DT v0.5 not working properly when DT used in shiny modules
```
library(shiny)
library(DT)

ui <- function() {
  TheModuleUI("module")
}
  
  callModule(TheModule, "module")
}

TheModuleUI <- function(id) {
  ns <- NS(id)
  fluidPage(
    DT::dataTableOutput(ns("table")),
    actionButton(ns("change"), "change first row")
  )
}

TheModule <- function(input, output, session) {
  data <- reactiveValues(df = data.frame(x = c(1,2), y=c('a','b'), stringsAsFactors = FALSE))

  output$table <- DT::renderDataTable({ DT::datatable(data = isolate(data$df)) })
  proxy <- DT::dataTableProxy("table")
  
  observeEvent(input$change, { data$df[1, "y"] <- 'changed' })
  observe({ DT::replaceData(proxy, data$df) })
}

shiny::shinyApp(ui = ui, server = server)
```

image

Then the example above is run on top of DT 0.4, then when clicked the "change" button, the "y" column in the first is changed properly to the new value due to reloading of updated data. In 0.5 the data is not reloaded.

result in 0.4 (first row changed):
image

result in 0.5 (first row not changed):
image

When I run the similar example without using shiny modules, everything works properly

donarus:
DT-626 - fixed issue with non-working reloadData/replaceData  method when called within Shiny Modules.

This bug was introduced in 820361cb1d917118b0b2551a5a9c69cb77e0a12b, where support for pomises was added.
The problem onlly occured when DT was used inside a shiny module. The reloadData used non-namespaced outputId
identifier and from this reason the initial callback URL differed from the URL on which the reloadData
binded.

This fixes #626

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -443,7 +443,7 @@ dataTableAjax = function(session, data, rownames, filter = dataTablesFilter, out
   data = as.data.frame(data)  # think dplyr
   if (length(rn)) data = cbind(' ' = rn, data)
 
-  sessionDataURL(session, data, outputId, filter)
+  sessionDataURL(session, data, session$ns(outputId), filter)
 }
 
 sessionDataURL = function(session, data, id, filter) {


jcheng5:
Thanks for this! And sorry for not investigating on Friday when I said I would.

While the fix looks very simple, I'm disturbed that the promise changes were what introduced the breakage--I'd like to understand that before we merge, in case there is some deeper problem going on here. Also, there's a new `shiny::getCurrentOutputInfo` function that might be able to greatly simplify this code, I wonder if it will also make the problem go away.

donarus:
Your PR looks much better - it solves the real cause and not only the result as mine. Thank you very much for a quick fix!!

PR #638, opened 2019-02-28
 Parameters to define custom filter function.

galachad:
This small change allows for defining a custom filter function in `renderDataTable`.

Currently, the only possibility to set the custom filter function is `dataTableAjax`, it's more clear to set the filter function in `renderDataTable`.


yihui:
Note that the argument name cannot be `filter`: https://github.com/rstudio/DT/commit/f4349373e7950b10de734bba57c6c050d708c5c4 I have renamed it to `filterFunc`.

PR #649, opened 2019-03-22
Add function to format as exponential

g1o:
FormatExp to always show a fixed number of digits in exponential format.


g1o:
I think i fixed the failed documentation test. So, I will close and try to pull it again.

PR #652, opened 2019-03-28
Add fuction to format as exponential

g1o:
The formatExp shows the number as exponential.  I hope it can pass the tests now.


R/format.R
```diff
@@ -93,6 +96,15 @@ formatSignif = function(
   formatColumns(table, columns, tplSignif, digits, interval, mark, dec.mark)
 }
 
+
+#' @export
+#' @rdname formatCurrency
```
yihui: Please add a brief description of this function on line 18 above.

R/format.R
```diff
@@ -93,6 +96,15 @@ formatSignif = function(
...
+formatExp = function(
+  table, columns, digits = 2, interval = 3, mark = ',', dec.mark = getOption('OutDec')
+) {
+  formatColumns(table, columns, tplExp, digits, interval, mark, dec.mark)
```
yihui: I don't think the arguments `interval`, `mark`, or `dec.mark` will work for `formatExp()`. You will have to either drop them or work harder on the `markInterval` function in JS.

inst/htmlwidgets/datatables.js
```diff
@@ -8,8 +8,9 @@
 var DTWidget = {};
 
 // 123456666.7890 -> 123,456,666.7890
-var markInterval = function(d, digits, interval, mark, decMark, precision) {
+var markInterval = function(d, digits, interval, mark, decMark, precision, exponential) {
   x = precision ? d.toPrecision(digits) : d.toFixed(digits);
+  if(exponential) {x = exponential ? d.toExponential(digits) : d.toFixed(digits);};
```
yihui: Here `d.toFixed(digits)` will never be executed no matter if `exponential` is true or false.

R/format.R
g1o: I don't know either, I would sure need to work harder on them, but most likely I will drop them (as I don't know what they do yet). I just copied the code and made a few changes to got it working.

PR #657, opened 2019-04-10
Added option to editable parameter for column selection

### LINKED ISSUE #550: Allow specific columns to be editable
Is there any way, to enable specific columns to be editable instead of whole datatable using either R or JS ?

opremicSebastian:
We implemented a small change that should close #550. Our solution extends the `editable` parameter of the `datatable` function to accept vectors specifying the editable columns.

In case there are any documentation things we missed (this is my first pull request on github ;-) ), please let me know and I will extend the pull request.

Cheers,
Sebastian

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -209,6 +209,7 @@ datatable = function(
 
   if (isTRUE(editable)) editable = 'cell'
   if (is.character(editable)) params$editable = editable
+  if (is.numeric(editable)) params$editable = jsonlite::toJSON(editable)
 
   if (!identical(class(callback), class(JS(''))))
     stop("The 'callback' argument only accept a value returned from JS()")
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -710,7 +710,16 @@ HTMLWidgets.widget({
     if (typeof data.callback === 'function') data.callback(table);
 
     // double click to edit the cell, row, column, or all cells
-    if (data.editable) table.on('dblclick.dt', 'tbody td', function(e) {
+    var editableSelector = 'tbody td';
+    if(Array.isArray(data.editable)) {
+      editableSelector = 'tbody td:nth-child(' + data.editable[0] + ')';
+      for (var i = 1; i < data.editable.length; i++) {
+        editableSelector += ', tbody td:nth-child(' + data.editable[i] + ')';
+      }
+      data.editable = 'cell';
+    }
+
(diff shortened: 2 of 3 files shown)


opremicSebastian:
Thank you for your comments @yihui! We started to look into the extensions you proposed. They seem to in be straight forward from what we already implemented and very useful.

During our first internal discussion, we came across a point we wanted to discuss with you: In your example how the parameters should look like, it seems to us that a (for us crucial) way to us this feature is dismissed:
Currently, in version 0.5.3 the option `editable = "column"` means that when clicking on a cell the complete column changes to input status. Do you want to transfer this to the example in your review? So that there is no possibility to edit only single cells while excluding columns from editing. From our point of view, we think that there is only the need to transfer the editing of single cells and neglect rows and columns. The new parameter `target="row"` (complete rows are excluded from editing) has nothing to do with the parameter `editable="row"` (the complete row changes to input mode) in the current implementation.

Thank you!

yihui:
Ah, I see. So you were looking for something like this:

```r
editable = list(target = 'cell', disallow = list(column = c(1, 3)))
```

i.e. you want to be able to edit cells, but disallow editing on cells in certain columns. Then the API is even more general, e.g.

```r
# all columns can be edited, but certain rows shouldn't be
editable = list(target = 'column', disallow = list(row = c(2, 5, 7)))

# all cells can be edited, but certain rows and columns shouldn't
editable = list(target = 'all', disallow = list(row = c(1, 3), column = c(3, 8, 10)))

# single cells can be edited except certain ones
editable = list(target = 'cell', disallow = list(cell = rbind(c(1, 3), c(2, 2), c(3, 5)))
```

opremicSebastian:
We included a first extension based on your comments. Currently, we decided to not implement the version I mentioned in my last comment, but the version you specified in your comment. It would be nice, if you can have a look at this new version and let us know what you're thinking about it. 

Actually, after seeing and testing the current implementation I don't know if it is necessary to make it more generalized. What do you think?

Best regards
Sebastian

yihui:
@opremicSebastian Thanks so much! I looked at your changes and felt it might be a little easier for me to maintain it if I do this by myself, so I did it via 832714f. I hope you don't mind. I only considered the case of disabling editing certain columns, though.

mazknio:
Is it possible to add a feature where specific rows and cells can also be restricted?

yihui:
@mazknio I don't have time to implement it. Sorry. I'd welcome pull requests.

PR #667, opened 2019-05-21
Escape mark and dec.mark for various formatters

### LINKED ISSUE #666: formatRound does not work for mark '
Similar to the problem with `formatCurrency()` in https://github.com/rstudio/DT/issues/135. The table below does not render.

```r
library(magrittr)
DT::datatable(data.frame(x = 1e5)) %>%
  DT::formatRound("x", mark = "'")
```

It seems as `'` needs to be escaped for this function as well (compare the fix in  cadb6df).

lorenzwalthert:
Closes #666 and is slightly more general in that it also fixes the problem not only for formatRound and also not only for the argument `mark`. The PR adds escaping for the arguments `mark` and `dec.mark` for the following formatters:

- `formatCurrency`
- `formatPercentage`
- `formatRound`
- `formatSignif`

### DIFF
--- a/R/format.R
+++ b/R/format.R
@@ -57,6 +57,7 @@ formatCurrency = function(
 ) {
   currency = gsub("'", "\\\\'", currency)
   mark = gsub("'", "\\\\'", mark)
+  dec.mark = gsub("'", "\\\\'", dec.mark)
   formatColumns(table, columns, tplCurrency, currency, interval, mark, digits, dec.mark, before)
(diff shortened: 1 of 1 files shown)


lorenzwalthert:
Well to move it inside `formatColumns()`?

yihui:
I meant the latter. It cannot be done in `formatColumns()` because you cannot modify `...` (at least not straightforward).

Issue #691, opened 2019-07-23
Columns escaped seem false in DT::datatable( ) with many columns


philibe:
Hello,

If I change  an initial column `myCol` to an url (for example), and copy the old column `myColInitialValue` at the end of the dataframe `df` with a new name, I thought that 
a `which(colnames(df)=='myCol')` send back the col # of `myColInitialValue` but It seems to be an issue in DT::datatable()

Here is a reprex below and a question in Stackoverflow:
- [dplyr - mutate: columns escaped are false in DT:datatable](https://stackoverflow.com/questions/57123813/dplyr-mutate-columns-escaped-are-false-in-dtdatatable-title-edited-was-dpl)

My goal is for the `escape` parameter of `DT::datatable()`. I use `escape=FALSE` in waiting that. With constants it doesn't work also but the [DT package](https://github.com/rstudio/DT/blob/master/R/datatables.R) seems also get the bad # column. :)

Here is my source with the issue of the bad column escaped:

- the # column is correct
- when I was debugging I get a dataframe with incorrect order of column 
but I didn't get again, I didn't reproduce it.
- but even with the correct number with `which()` the escaped column displayed in shiny/ datatable is wrong

I cannot upgrade all packages because I have qualification server and production server but not really test computer in my small company, I cannot break all. 

```R

options(encoding = "UTF-8")

library(DT)
library(shiny)
library(dplyr)
library(stringr)

hostipserver <- str_trim(system("hostname -I", intern=TRUE))
hostportserver <- ":8080"

app<-
  shinyApp(
    ui = basicPage(
      navbarMenu("Bla",
                 tabPanel("blabla",
                          fluidPage(
                            h3("outblabla_1"),
                            p("toto_1 and toto_2 have to be worked urls but only toto_2 is ok. varcible is a worked url but I don't want it."),
                            fluidRow(
                              column (12,
                                      div(DT::dataTableOutput('outblabla_1'), 
                                          style = "font-size:80%;white-space: nowrap;width:93%")
                              )
                            ),
                            h3("outblabla_2"),
                             p("toto_1 and toto_2 have to be worked urls but only toto_2 is ok"),
                            fluidRow(
                              column (12,
                                      div(DT::dataTableOutput('outblabla_2'), 
                                          style = "font-size:80%;white-space: nowrap;width:93%")
                              )
                            )                            
                          )
                 )
      )           
    ),
    
    server = function(input, output) {
      
      blabla <-  reactive({
        test<-data.frame(        
[...]

PR #693, opened 2019-07-23
doc should increase numeric column indices by one when row names are displayed

shrektan:
Referred to #691 

Since there're tons of files being updated if I rebuild the site, I only update the Rmd file in this PR.


yihui:
FYI I actually run `make site` to build the site. I have tons of hacks in the build script...

PR #702, opened 2019-08-22
`formatStyle()` with `styleEqual()` is now chainable by default

### LINKED ISSUE #632: Multiple formatStyle with target='row' not working
This issue was originally described in:

https://github.com/rstudio/DT/issues/202

I believe that results shown in the response closing the ticket removed the target='row' statement.  When I remove it I get the results shown, however, when it is present only the first formatStyle is applied.  Is it possible to use two different formatStyle statements to format rows based on different columns such as noted in the original post:

```r
library(DT)
myTable = datatable(
  data.frame(V1 = rep(c('my1', 'my2'), 6), V4 = rep(c('my2', 'my1'), each = 6))
)
myTable %>% formatStyle(
  'V1',
  target = 'row',
  backgroundColor = styleEqual(c('my1'), c('#ea9f9f'))
) %>% formatStyle( # doesn't work #
  'V4',
  target = 'row',
  backgroundColor = styleEqual(c('my2'), c('#e8cf9d'))
)

```

shrektan:
Closes #632 

Although the default value of `default` is changed, I believe the current default behavior (left the non-matched cells' CSS value unchanged) is almost always desired.

## Example

```r
library(DT)
myTable = datatable(
  data.frame(V1 = c('a', 'b', 'c', 'd'), V2 = c('1', '2', '3', '4'))
)
myTable %>% formatStyle( 
  'V1',
  target = 'row',
  backgroundColor = styleEqual('a', 'red')
) %>% formatStyle(
  'V1',
  target = 'row',
  backgroundColor = styleEqual('c', 'green')
)
```

### Before the PR 

(Note the first row's color is blank, meaning it gets overridden by the 2nd `formatStyle()`)

image

### After the PR

(Works as expected)

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.8.3
+Version: 0.8.4
 Authors@R: c(
     person("Yihui", "Xie", email = "xie@yihui.name", role = c("aut", "cre")),
     person("Joe", "Cheng", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # CHANGES IN DT VERSION 0.9
 
+## NEW FEATURES
+
+- `formatStyle()` with `styleEqual()` is now chainable by default (thanks, @e-kennedy #632).
+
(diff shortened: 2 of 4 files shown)


philibe:
I've seen it but I thought it was a feature to not have two similar style on same target. Good news. :)

PR #717, opened 2019-10-06
Integrate Select Extension

### LINKED ISSUE #366: Shiny rows_selected does not react to selectAll Button
Hi,

It seems the shiny value input$...._rows_selected does not react to the selectAll button from the Select extension. Clicking the "Select All" button correctly selects the rows in the table, but the selection is not sent to Shiny. However, once unselecting a row, the rest of the selection is correctly sent to Shiny.

A minimal example:

```
library(shiny)
library(DT)

ui <- fluidPage(

  title = 'Select Table Rows',

  h1('A Server-side Table'),

  fluidRow(
    column(9, DT::dataTableOutput('x3')),
    column(3, verbatimTextOutput('x4'))
  )

)

  # server-side processing
  mtcars2 = mtcars[, 1:8]
  output$x3 = DT::renderDataTable(datatable(mtcars2,options = list(dom = 'B',buttons=c("selectAll","selectNone")),extensions=c("Select","Buttons")), server = TRUE)

  # print the selected indices
  output$x4 = renderPrint({
    input$x1_rows_selected
    s = input$x3_rows_selected
    if (length(s)) {
      cat('These rows were selected:\n\n')
      cat(s, sep = ', ')
    }
  })

})
shinyApp(ui,server)
```

Thanks,

Andras

### LINKED ISSUE #572: Shift + Click not working with cell selection (Shiny)
In a data table that supports the selection of individual cells, Shift + click will not select all cells between the two clicked ones, but rather just mark the values and only select the two cells that were clicked on.
```
shinyApp(
  ui = fluidPage(DTOutput('tbl')),
  server = function(input, output) {
    output$tbl = renderDT(
      datatable(iris, selection=list(target="cell"))
    )
  }
)
```

### LINKED ISSUE #706: Integrate Select extension
Now datatables library has the official Select extension. Due to historical reasons, DT implements the selection functionality separately. Replacing the current selection implementation with Select extension can be beneficial:

1. Fewer conflicts with other extensions;
1. Reduce javascript codes so easier to maintain;
1. Consistent styles - like bootstrap or other table themes.

I have wished for this for a long time... Took a look at the last weekend and it should be technically possible...

shrektan:
Working in progress.

Closes #706 Closes #366 Closes #572

- [x] Row/Column selection in Client mode
- [x] Cells in Client mode
- [x] Row/Column/Cell's existing selection info can be kept in both client and server mode after re-drawed (e.g., switch pages)
- [x] 'single' style works expectedly for both client and server mode
    - [x] [fixed] Except for one case that is if you select row 1 on page A, switch to page B then switch back, click on row 1 again won't deselect it. This is because the Select extension can't properly handle this and trigger the select events again (should be the deselect event). I think we can improve this by comparing to the existing selections first (before clean up the selection array)
- [ ] The server processing mode
    - On server mode, the `os` style (multiple selections) is different from current behavior. More specifically, [the cross-page selection using Shift is not possible in DT's Select Extension](https://datatables.net/forums/discussion/37781/select-extension-loosing-selection-with-server-side-paging-or-scroller-extension). I can't see there's a way to use Select extension with Shift to select rows across pages.
    - On server mode, the `deselect` event will not be triggered so we need some code to handle this differently from the client mode. 
        - It seems difficult or impossible without considering what users have keyed. 
        - For example, if the user doesn't press the `Ctrl` button, we should `deselect` the pre-selected rows. If not, we shouldn't.
        - This means we need to re-implement all the Select logic for the server-side processing mode by ourselves, inevitably. [The source code of Select ext.](https://cdn.datatables.net/select/1.3.1/js/dataTables.select.js) is a good place to start with.
        - ~~The whole point is, all the valid clicks will end up triggering the events `select` or `deselect`. We just need to record the last clicked cells and override the selection range.~~(not correct)
        - There's another issue that is the event object `e` passed to the `select/deselect` event seems to contain no info about what the user has keyed. We may have to register the handling on the `click` events instead of `select/deselect` event.
- [ ] Auto-Convert the old options to the current implementations
- [ ] How to handle the historical code that uses Row+Column features?
- [ ] should handle the `API` selection, a.k.a., using buttons to select should trigger the events.
    - It may be problematic for the server mode. ~~I think the `Select All` buttons will not trigger any specific events but select all rows in the current page. Any way to detect that so we can implement a server logic to handle the event properly?~~(see the next bullet)
    - It's possible by adding an event about the `Select All` and `Select None` buttons. Overriding the selection when it's server-processing mode.
- [ ] Be sure it works expected when filtering
- [ ] Be sure it works fine when Select extension is not in use

## Issues to be fixed

- [ ] server = TRUE, style = 'multi': select the row 1 on page 1, switch to page2, switch back to page1, you have to click twice on row 1 in order to deselect it. It's because the Select extension is unaware of row1 has been selected.

## Code to test

Should test both the client and server mode; should test all the style and items combinations; for multiple selections, should test the cross-page functionality

```r
library(shiny)

ui = fluidPage(
  radioButtons('server', 'server', c('TRUE', 'FALSE'), inline = TRUE),
  radioButtons('style', 'style', c('os', 'multi', 'single', 'multi+shift'), inline = TRUE),
  radioButtons('items', 'items', c('row', 'column', 'cell'), inline = TRUE),
  DT::DTOutput('tbl'),
  verbatimTextOutput('print')
)
server = function(input, output) {
  output$print = renderPrint({
    print(input$tbl_rows_selected)
    print(input$tbl_columns_selected)
    print(input$tbl_cells_selected)
  })
  output$tbl = DT::renderDT(DT::datatable(
    selection = 'none', 
    iris, extensions = c('Select', 'Buttons'), options = list(
      select = list(style = input$style, items = input$items),
      dom = 'Blfrtip',
      rowId = 0,
      buttons = c('selectNone', 'selectRows', 'selectColumns', 'selectCells')
    )
  ), server = as.logical(input$server))
}
runApp(list(ui = ui, server = server), port = 5759)
```


shrektan:
After re-read the source file of the Select extension, now I don't think we can really integrate the server-side selection using the Select extension as planned. 

The reason is that all the Select related API implementations are now only suitable for the client-side mode. There's no good interface to handle the server-side processing mode at all. Hacking all the possible cases is difficult and may result in a messy.

So before a direct interface for server-side processing is provided in the Select extension, we should:

1. For the client mode, we should support both the current `selection` option and the Select extension. If both are enabled, we should warn that only the former is actually used. 
1. For the server-side processing mode, we should throw errors or warnings if the Select extension is enabled. 
1. We should improve our own implementation of selection to be as close as the Select extension. That means we should support Ctrl (now only Shift is supported). We should support the ctrl/shift for columns and cells as well. In addition, the selection style class should be configurable or aligned with the themes.

PR #727, opened 2019-11-13
Use Perl REs in filters

rfhb:
This PR is to extend the filtering by allowing extended regular expressions, which users will want to use for example to look ahead or behind, or to negate search terms.


rfhb:
Many thanks for considering and reviewing!

PR #730, opened 2019-11-20
Use current outputId in dataTableAjax() if available

alandipert:
We noticed when testing [shinyloadtest](https://github.com/rstudio/shinyloadtest) that the shiny example [063-superzip-example](https://github.com/rstudio/shiny-examples/tree/master/063-superzip-example) didn't work.

It turned out the reason for the failure was non-determinism in the application. The example used `DT::dataTableAjax()` and [didn't supply an `outputId` parameter](https://github.com/rstudio/shiny-examples/blob/918cd7971353982e5f5256102b6b96a8a31959b5/063-superzip-example/server.R#L179), which caused DT to generate a random one using `tempfile()`.

We were able to fix the example by [simply specifying an outputId](https://github.com/rstudio/shiny-examples/pull/179), but it occurred to us that we could maybe prevent others from running into this problem by making a small enhancement to DT.

@jcheng5 pointed out that since `dataTableAjax()` was first written, Shiny (in 1.2.0) acquired the function `getCurrentOutputInfo()`, which can be used to obtain the name of the current output. This seems like a better default.

I wasn't sure if `dataTableAjax()` would always be called in a dynamic environment in which `shiny::getCurrentOutputInfo()[["name"]]` was non-`NULL`, so I added a NULL check and retained the `tempfile()` generation for that case.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -448,7 +448,9 @@ shinyFun = function(name) getFromNamespace(name, 'shiny')
 #'   list(value = 'FOO', regex = 'false'), length = 10, ...)}) that return the
 #'   filtered table result according to the DataTables Ajax request
 #' @param outputId the output ID of the table (the same ID passed to
-#'   \code{dataTableOutput()}; if missing, a random string)
+#'   \code{dataTableOutput()}; if missing, an attempt to infer it from
+#'   \code{session} is made. If it can't be inferred, a random id is
+#'   generated.)
 #' @references \url{https://rstudio.github.io/DT/server.html}
 #' @return A character string (an Ajax URL that can be queried by DataTables).
 #' @example inst/examples/ajax-shiny.R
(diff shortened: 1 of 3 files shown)


R/shiny.R
```diff
@@ -457,8 +457,9 @@ dataTableAjax = function(session, data, rownames, filter = dataTablesFilter, out
 
   oop = options(stringsAsFactors = FALSE); on.exit(options(oop), add = TRUE)
 
+  if (missing(outputId)) outputId <- shiny::getCurrentOutputInfo()[["name"]]
```
wch: Better to use `session$` instead of `shiny::` here.

wch:
Since the ajax-shiny.R app calls `dataTableAjax()` outside of an output, where `getCurrentOutputInfo()` would return nothing, it would be good to change that example to provide an `outputId`.

Also, the documentation for `outputId` needs to be updated to reflect the new behavior.

PR #738, opened 2019-12-11
should use scroller = TRUE to initialize Scroller extension

shrektan:
From https://cdn.datatables.net/scroller/2.0.0/ we know that using `dom = 'S'` is deprecated after Scroller version 2.0.0 

> Note also that the option to initialise Scroller through the dom option (with the S feature) has been removed, and scroller should be used instead.


yihui:
Thanks!

PR #740, opened 2019-12-17
Add datatableThemeVariables() for setting datatable's SASS variables

cpsievert:
Adds the ability to control [datatable's SASS variables](https://github.com/DataTables/DataTablesSrc/blob/8ea4882d84ea94c016efdfcc70b14f0ac4207b66/css/jquery.dataTables.scss#L2-L36) through a new function `datatableThemeVariables()` (the implementation of which is basically a copy/paste job of `bootstraplib::theme_variables()`). This should come in handy for both users and developers (i.e., packages like rmarkdown, flexdashboard, and shiny could/should use Bootstrap/Bootswatch SASS variables to inform datatable's SASS)


cpsievert:
BTW, I'm thinking `rmarkdown::html_document_base()` will [do this](https://github.com/rstudio/rmarkdown/pull/1706/commits/e80f39546f5024c0ab59623e92d658f683f86df4) to set the `bootswatch.theme` option (feel free to suggest a better name)(e80f39546f5024c0ab59623e92d658f683f86df4) and `bs4::bs4_page()` will eventually do something similar

cpsievert:
Note to self: if new `getCurrentOutputInfo()` styles are available, those should take precedence

R/datatables.R
```diff
@@ -618,6 +627,16 @@ depName = function(style = 'default', ...) {
   tolower(paste(c(..., if (style != 'default') c('-', style)), collapse = ''))
 }
 
+styleNormalize = function(style) {
+  style = tolower(style)
+  if (identical(style, 'auto')) {
+    if (!length(bsThemeGet())) return('default')
+    version = bootstraplib::theme_version()
+    style = if (version %in% "3") "bootstrap" else if (version %in% c("4", "4+3")) "bootstrap4"
```
cpsievert, line 635:
```suggestion
    style = if (version %in% "3") "bootstrap" else "bootstrap4"
```

cpsievert:
I hadn't realized that compilation of datatables Sass isn't even relevant for Bootstrap styling, only for the `default` styling approach. 🤦 

As a result, I might drop this idea entirely (in favor of #852), because it adds quite a bit of complexity without much added benefit

PR #751, opened 2020-01-06
Feature/complex headers

mtyszler:
Fixes rstudio#418.

This version builds on original suggestions from @chrisvwn.

It creates a function that reads multi-row headers into an array of arrays, where each arrary row is a header row. 

Further it updates csv, copy, pdf, excel and print to process the multi-row header object.

For excel it merges the cells as shown on screen.

It does not update the flash counter parts of html5 functions

### DIFF
--- a/inst/htmlwidgets/lib/datatables-extensions/Buttons/js/buttons.html5.min.js
+++ b/inst/htmlwidgets/lib/datatables-extensions/Buttons/js/buttons.html5.min.js
@@ -444,14 +444,112 @@
             p.title && (r([p.title], g), k(g, /*d.header.length*/ d.header[d.header.length - 1].length - 1));
             p.messageTop && (r([p.messageTop],
                 g), k(g, /*d.header.length*/ d.header[d.header.length - 1].length - 1));
+
             /* ----- BEGIN added Code ----- */
             //c.header && (r(d.header, g), i("row:last c", e).attr("s", "2"));
 
 			if(c.header){
+				row_start_headers = g;
 				for (o = 0; o < d.header.length; o++){
 	                  r(d.header[o], g);
 	                  i("row:last c", e).attr("s", "2");
 	                }
+            }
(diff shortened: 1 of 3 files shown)


shrektan:
@yihui  I share the same concern, too. These are not simple hacks. Even if we find out a way to upgrade the js files with this patch maintained (one possible way is to apply a git patch), I can't guarantee it will always work in the future because the upstream source may contain big changes.  

I incline that this PR is more suitable for [the upstream js library](https://github.com/DataTables/Buttons) but you have more experience on this topic.

yihui:
Absolutely. If this could be done in the upstream library instead, we can easily pick it up from there in the future.

mtyszler:
Thank you @yihui and @shrektan for your review and comments. I wasn't aware that the code was auto-imported via script and I understand your concerns. 

In the upstream library there is an open PR (https://github.com/DataTables/Buttons/pull/55) with similar topic. However, it has been open over 5 years ago with last comment (https://github.com/DataTables/Buttons/pull/55#issuecomment-540957219) from @Datatables is a bit ambiguous in how they want to proceed. 

I'm not sure I have the time now to reproduce my solution at a fork of https://github.com/DataTables/Buttons, but if I do I might try. 

I'll ask @Datatables if they would be interested in that before proceeding. 

So I take you'd prefer not to merge this PR, right?

yihui:
That's right.

The worst case is that if users really want this, we provide an R script or function to download the patched version of these JS files and replace the ones in **DT**. Of course, we should also warn them that the next time they update **DT**, they should run the R script or function again.

mtyszler:
Dear @yihui and @shrektan.

I just submitted https://github.com/DataTables/Buttons/pull/170 to the upstream library, as you suggested.

I suggest you keep this PR open and close it when/if upstream changes are taken up and picked-up by you.

Best regards,

PR #754, opened 2020-01-17
add scrollResize plugin

### LINKED ISSUE #467: ScrollResize plugin
I'm looking to use the "ScrollResize" plugin for datatables (https://github.com/DataTables/Plugins/commit/9bae300872876edf69c90bb0854d555fcfb64f58).

This can be useful when using paging=FALSE and instead of specifying scrollY=200, specify instead scrollResize=T to automatically calculate the height to fill the container the datatable lives in.

Hope that this isn't too much work, and I'm not sure what it would take but I could try to look into it if I find the time too.

shrektan:
Closes #467 

See https://datatables.net/blog/2017-12-31 for more info about this plugin.

### Example code

Try to resize the red area. With this plugin, the table will fit it well. Otherwise, it will be fixed to a certain height.

```r
library(shiny)
ui <- tags$html(
  tags$head(
    tags$link(
      rel = "stylesheet", href = "//code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css"
    ),
    tags$script(
      src = "https://code.jquery.com/ui/1.12.1/jquery-ui.js"
    ),
    tags$script(
      DT::JS("$(function() {$('#resizable').resizable();});")
    )
  ),
  tags$body(
    radioButtons('scrollresize', 'scrollresize', choices = c('TRUE', 'FALSE')),
    div(
      id = 'resizable',
      style = 'height:500px;background-color:red;',
      DT::DTOutput('tbl', height = '100%')
    )
  )
)

  dat <- iris[1:50,]
  output$tbl <- DT::renderDT(
    if (input$scrollresize == 'TRUE') {
      DT::datatable(
        dat, plugins = 'scrollResize',
        options = list(scrollResize = TRUE, paging = FALSE, scrollY = '100px', scrollCollapse = TRUE)
      )
    } else {
      DT::datatable(dat, options = list(paging = FALSE, scrollY = '100px', scrollCollapse = TRUE))
    }
  )
}

shinyApp(ui, server)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.11.2
+Version: 0.11.3
 Authors@R: c(
     person("Yihui", "Xie", email = "xie@yihui.name", role = c("aut", "cre")),
     person("Joe", "Cheng", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - Support the Select extension on the client-side processing mode (thanks, @shrektan, #744).
 
+- Support [features/scrollResize plug-in](https://datatables.net/blog/2017-12-31) (thanks, @andrewkho @shrektan, #467 #754).
+
 ## BUG FIXES
 
 - The server no longer throws warnings when conducting searching (thanks, @bquast @shrektan, #749).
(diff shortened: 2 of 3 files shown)


aleaficionado:
I'm struggling with the example above. All it seems to do is fill the table to a fixed size 500px which seems to be the same behaviour as scrollY=500px. Is there a way to get it to fill the whole body?

Issue #333, opened 2016-08-02
Can I focus search box and place cursor in end of preset text?


dracodoc:
I want to bring up a shiny gadget when some text was selected, and the gadget has a DT with search box preset with the selected text. 

@yihui , Can I focus the search box and place cursor in end of preset text? Thus I can continue to input or modify preset text.

I searched around a lot, only found some places mentioned use JS calls on independent widget like text area, not for search box of DT.

yihui:
Yes you can. Please see http://rstudio.github.io/DT/007-search.html

dracodoc:
@yihui , sorry I read that page several times but didn't find how to focus the cursor. The search box is filled with the text, but not focused.

yihui:
Oh, sorry, I missed that. Here is an example:

```r
library(DT)
mtcars2 = mtcars[, c(1:5, 9)]
datatable(
  mtcars2, colnames = c('model' = 1),
  options = list(
    search = list(search = 'M'),
    pageLength = 5
  ),
  callback = JS("$(table.table().container()).find('input').first().focus();")
)
```

dracodoc:
@yihui Ah, I guessed it's some JS magic that I don't know about it.

Though I'm using DT in a RStudio Addin with shiny gadget, so I have to create a `datatable` first then transfer to `renderDataTable`. I'm not sure if my code is optimal, at least it works now.

BTW, [my addin is here](https://github.com/dracodoc/namebrowser) in case you would want to check it out.

Issue #563, opened 2018-06-22
DT columns don't span width of widget


jayhesselberth:
I have a DT widget in a pkgdown website:

https://jayhesselberth.github.io/widgetdown/articles/dt.html

The widgets span the entire container, but the columns do not. Have tried setting `fillContainer`, `autoWidth`, and both.

Ideas for how to fix? Sorry if this is dup.

jayhesselberth:
This might help.

https://github.com/rstudio/flexdashboard/blob/dd4440a0424dc86c265af6af5a27b5bc912311be/R/flex_dashboard.R#L177-L183

jayhesselberth:
I have tried applying some custom [CSS from flexdashboard](https://github.com/rstudio/flexdashboard/blob/master/inst/rmarkdown/templates/flex_dashboard/resources/flexdashboard.css#L659-L678) and still see the issue.

I have tried using `fillContainer = TRUE` in the call, and it just squishes the container:

https://jayhesselberth.github.io/widgetdown/#dt---fillcontainer

jayhesselberth:
Column widths are fixed by adding `filter = 'top'` or `filter = 'bottom'`. Seems like the extra CSS included with the `filter` option (`selectize` and `nouislider`) may help.

https://jayhesselberth.github.io/widgetdown/articles/dt.html#dt---saurfang-example

shrektan:
It's really strange. However, I haven't used pkgdown before. Would you please provide a minimal reproducible example so that we can debug? Thanks.

jayhesselberth:
Here's a minimal package:

https://github.com/jayhesselberth/widgetdownreprex

And the rendered vignette demonstrating the issue in the site:

https://jayhesselberth.github.io/widgetdownreprex/articles/test.html

The relevant file is `vignettes/test.Rmd` in the repo. 

Clone the repo and run `pkgdown::build_site()` to get a preview of the site.

shrektan:
Sorry for the very late reply. I can still reproduce this. It's super weird for me.

I'll take some time fixing this.

Thanks for the reporting.

shrektan:
The cause of this issue is the custom css file defined by pkgdown:

```css
table {
  display: block;
  overflow: auto;
}
```

The table's display attribute is set to `block`... I'm not sure how to fix this. Maybe we should always set DT's table.display to "table" but I'm afraid of side-effects (should be none).

PR #767, opened 2020-02-01
Explicitly set the datatables' color to black

### LINKED ISSUE #447: DT table contens are invisible: RStudio + dark editor themes
I use RStudio and its editor's theme "Cobalt".

In R Markdown document I created a `DT::datatable`. Unfortunately, the contents were invisible:
invisible contents

The contents existed, but the font was as white as the background:
invisible contents highlighted

How can I use `DT::datatable` with dark editor themes in RStudio and get visible results?

### LINKED ISSUE #781: DT prints "blank" table in RStudio in certain editor themes
Working in RStudio 1.2.5033 with the built-in Monokai theme. 

I ran  the "hello world" for `DT`: 

```
library(DT)
datatable(iris)
```

and saw an empty table:

Screen Shot 2020-03-17 at 8 44 56 PM

but when I ran my mouse over it I saw the table values were printed in white:

Screen Shot 2020-03-17 at 8 45 13 PM

When I changed the background to Textmate the table printed fine:

Screen Shot 2020-03-17 at 8 57 47 PM

I get the same behavior with the CRAN and GitHub versions of the package. Does `DT` only work with certain editor themes in RStudio?

shrektan:
Fixes #447
Fixes #781

**UPDATE3 - 200318**

The upstream said they will have this feature in the v2 branch. Then we can revert this hotfix.

https://github.com/DataTables/DataTablesSrc/issues/157#issuecomment-600533021

---

**UPDATE2 - 200318**

After a "third" thought, despite I think **the final solution will better be using the dynamic rendering approach, as @cpsievert suggested,** this hotfix should be harmless to merge. From the screenshot we can know that **all the other font elements of datatables' default theme are using `#333` as the font color**.  So, in my opinion, there's no obvious reason that the font color of the table's content should be different. 

Anyway, I've filed an issue on the upstream https://github.com/DataTables/DataTablesSrc/issues/157, let's see the response first.

---

**UPDATE** After a second thought, it's not correct that all the theme text is black - it depends on the theme of the page. For example, if we use any bootswatch theme like flaty, the text is close to black but not exactly black. I need to re-think this. 

---

The datatables' theme CSS file doesn't define the color but with "white" background. It leads to an issue that when the HTML's body color is set to 'white', the user can't see the text since the background is white. **One case happens in the RStudio's IDE with theme "Cobalt", when inline viewing the DT table inside an Rmd file.**

This fix should have no side-effects because all the datatables' themes use the black text actually. Even if it doesn't, the theme will have to explicitly define the color and it should override this CSS definition.

### BEFORE 

截屏2020-02-02上午5 20 10

### AFTER

截屏2020-02-02上午5 20 32


cpsievert:
@shrektan please have a look at #740, which should provide a general way to fix issues like this

shrektan:
@cpsievert I believe being able to dynamically change the CSS attributes would be the solution for both #447 and #563. However, I'm not very clear on how it's possible to apply the CSS changes within different contexts. 

I mean, for #447, I would like to apply the `background: transparent` or `color: black` only when rendered in RStudio IDE. For #563, apply a `table.dataTable { display: table; }` only when it's rendered by `pkgdown`. It's so that I can ensure it won't cause trouble in unexpected situations.

However, I think we may need `htmlwidgets:::print.htmlwidget()` to have a callback or something. 

Is my understanding correct?

Thanks.

cpsievert:
> However, I think we may need htmlwidgets:::print.htmlwidget() to have a callback or something. Otherwise, I don't know how we can be aware of the rendering context...

Yea, that's essentially what's happening here (`preRenderHook()` gets called prior to the print method) https://github.com/rstudio/DT/pull/740/files#diff-047808c2d95cfa74750ef48ac75ad541R297-R316

> I mean, for #447, I would like to apply the background: transparent or color: black only when rendered in RStudio IDE.

I would wait on tackling this until #740 is merged, then set defaults for the relevant Sass variables. This way you'd still allow users to override those defaults in R.

shrektan:
(First of all, please see my last two updates in the first comment)

@yihui I think we may merge this as a hotfix since it affects many users. Due to the fact that the background is fixed to "white" and the font color of all the buttons is fixed to `#333`, I think fixing the font-color of the body is not a very bad choice. I personally believe the impact should be very minimal.

In the future, the upstream may change these inconsistent CSS styles (see https://github.com/DataTables/DataTablesSrc/issues/157#issuecomment-600533021), which means both the background-color and font-color will be relative values. At that time, we can reverse this patch.

Nevertheless, I'll let you make the final decision.

Thanks.

cpsievert:
@shrektan @yihui mainly commenting now so I remember to investigating reverting this change in #740 (and potentially using `rstudioapi::getThemeInfo()` to inform the foreground/background color)

PR #788, opened 2020-03-27
More tests

shrektan:
Just add more tests with very little code change when adding the tests.

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -361,7 +361,7 @@ convertIdx = function(i, names, n = length(names), invert = FALSE) {
 
 #' @importFrom htmltools HTML htmlEscape
 escapeData = function(data, i, colnames) {
-  if (is.null(data) || prod(dim(data)) == 0 || identical(i, FALSE)) return(data)
+  if (is.null(data) || prod(dim(data)) == 0 || isFALSE(i)) return(data)
   i = convertIdx(i, colnames, ncol(data))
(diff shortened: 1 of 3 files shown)


shrektan:
I'll merge this by myself as they are all about tests.

R/datatables.R
```diff
@@ -361,7 +361,7 @@ convertIdx = function(i, names, n = length(names), invert = FALSE) {
 
 #' @importFrom htmltools HTML htmlEscape
 escapeData = function(data, i, colnames) {
-  if (is.null(data) || prod(dim(data)) == 0 || identical(i, FALSE)) return(data)
+  if (is.null(data) || prod(dim(data)) == 0 || isFALSE(i)) return(data)
```
yihui: `isFALSE()` was introduced in R 3.5.0 (see [NEWS](https://cran.r-project.org/doc/manuals/r-release/NEWS.html)), so this means DT will require R >= 3.5.0. I'd prefer not increasing the R version requirement.

shrektan: It's not the `base::isFALSE()`. It's the util function defined by DT.

https://github.com/rstudio/DT/blob/431b7864861887fac993326e31edaa6cb2ffca14/R/utils.R#L6

yihui: Oh I see. Apparently I often forget functions that I wrote in the past 😂

Issue #462, opened 2017-10-26
Update filter ranges when the data is updated via replaceData


tm3da:
Hi,

Is there a way how to update filter ranges when I update DT using replaceData? 

Example:
When you click on the button it changes value in 3rd row to 10 000 but filter ranges only from ~4 to ~8. (I used the Iris dataset for the example).
 
```
library(shiny)
library(DT)

ui <- fluidPage(
    actionButton("chbtn", "change sepal length to 10 000"),
    dataTableOutput("dt")
)

server <- function(input, output, session){
    data <- iris

    output$dt <- renderDataTable({
        datatable(data, filter = "top")
    })

    observeEvent(input$chbtn, {
        dtp <- dataTableProxy("dt")
        data[3, 1] <- 10000
        replaceData(dtp, data, resetPaging = FALSE, clearSelection = "none")
    })
}

shinyApp(ui, server)
```

Screen shot:
![image](https://user-images.githubusercontent.com/1069014/32044273-6a7e5662-ba3d-11e7-9f60-e85ef5e48763.png)

paulmessinesi:
Hi, I came across this too. Documentation is clear but is there a way to force the filter options though? For instance, feeding a list of possible factors one could add to a specific column when the table is first rendered before any "replaceData" happens? Thanks for your help and the great package.

shrektan:
No, I'm afraid it's not possible right now.

Mosk915:
I was able to accomplish this by writing a javascript function and calling it via ShinyCustomMessageHander within a wrapper function for replaceData.

```javascript
Shiny.addCustomMessageHandler("updateFilters",
  function(x) {
    // outputId of the DT element
    var table = document.getElementById(x.tableid);
    // filter row of the table element within the DT element
    var tablerow = table.getElementsByTagName('table')[0].rows[1];
    // loop through each cell in the filter row
    for (var i = 0, cell; cell = tablerow.cells[i]; i++) {
      // get list of all data elements for the column from the data going into the table (duplicates are okay)
      var items = x.data[Object.keys(x.data)[i]];
      // for factor columns
      if(['factor', 'logical'].includes(cell.getAttribute('data-type'))) {
        // find the selectize object
        var dropdown = $(cell).find('.selectized').eq(0)[0].selectize;
        // get the current value of the selectize object
        var curr_value = dropdown.getValue();
        // put the list of data elements in the appropriate format for the selectize
        items = items.map(function(item) {return {text: item, value: item}});
[...]

PR #789, opened 2020-03-31
Use the endpoints of the noUiSlider for filtering

Mosk915:
There is no built in method to update the slider and dropdown filters when data is replaced (#462), however, it is possible for these widgets to be updated using javascript.

In order for the slider to work correctly, the minimum and maximum values of the column need to be based on the endpoints of the slider, instead of hardcoded based on the initial data.


Mosk915:
Glad I could contribute.

PR #800, opened 2020-04-21
uses GitHub action

shrektan:
Use Github Action to run R CMD check and test coverages.

The only advantage, for now, is that if something goes wrong we can just see it inside of Github's Action button... Another (not very convincing) reason is that the Github Action seems promising...

I admit it's not useful so feel free to close this PR if it's not necessary.

Thanks 😄


.github/workflows/r-cmd-check.yaml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  R-CMD-check:
+    runs-on: macOS-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: r-lib/actions/setup-r@master
+      - uses: r-lib/actions/setup-pandoc@master
+      - name: Query dependencies
+        run: |
+          install.packages(c('remotes', 'rcmdcheck', 'covr', 'rmarkdown'))
```
yihui, line 14: Two questions:

1. Can these packages be cached so they don't need to be installed every time?

1. Are there any obvious advantages of using the **rcmdcheck** package compared to the simple command line `R CMD check`? I have to admit that I'm not familiar with this package.

.github/workflows/r-cmd-check.yaml
```diff
@@ -0,0 +1,35 @@
...
+          remotes::install_deps(dependencies = TRUE)
+          saveRDS(remotes::dev_package_deps(dependencies = TRUE), "depends.Rds", version = 2)
+        shell: Rscript {0}
+      - name: Cache R packages
+        uses: actions/cache@v1
+        with:
+          path: ${{ env.R_LIBS_USER }}
+          key: macOS-r-3.6-${{ hashFiles('depends.Rds') }}
```
yihui, line 22: Is this for R 3.6.x only? I wonder how caching for multiple versions of R works. For Travis, it is fairly simple: https://github.com/rstudio/DT/blob/7ebd4cc472e8d08da892a4bacd95b8ac1245ab48/.travis.yml

.github/workflows/r-cmd-check.yaml
```diff
@@ -0,0 +1,35 @@
...
+        shell: Rscript {0}
+      - name: yihui gitconfig
+        run: |
+          curl https://xran.yihui.org/.gitconfig -o ~/.gitconfig
+      - name: Test coverage
+        run: covr::codecov()
+        shell: Rscript {0}
+      - name: xran
+        run: |
+          curl https://xran.yihui.org/r-xran | bash
```
yihui, line 35: For this to work, I need to set [an environment variable containing my Github token](https://github.com/yihui/xran/blob/ad12fcfc1e89aea75e7f3f8270b73cca274965cd/r-xran#L12). How do I encrypt that environment variable?

yihui:
> To conclude, I can work on doing updating this PR in DT with lastest change in actions if you want.

Please feel free to go ahead and do it. Thanks!

cderv:
Closing this as we merged #873

PR #873, opened 2020-11-03
Switch to Github actions

cderv:
This replace #800 by updating to last convention for https://github.com/r-lib/actions

Setup using 
* `usethis::use_github_action_check_standard()` 
* `usethis::use_github_action("test-coverage")`
* 

Modification: 
* Use ubuntu 16-04 like on travis - should we update ? 

Not included yet: 

- [ ] Publishing on xran. I'll try basic support in the workflow but we could make a specific action for this I believe.


cderv:
So this works quite fine an pretty quick. 

Just the part about xran need to be checked and maybe re think as made for travais in the first place.

cderv:
@yihui @shrektan before merging, what do you think ? 
@shrektan this would replace your previous PR.

cderv:
I am merging this now - switching to GHA officially

PR #876, opened 2020-11-10
Add support for shiny::bindCache()

wch:
With this simple example app, the first user will see a delay before it shows the table, but other users who visit the app should see it instantly.

```
library(shiny)
library(DT)
shinyApp(
  ui = fluidPage(DTOutput('tbl')),
  server = function(input, output) {
    output$tbl <- renderDT({
        message("Computing...")
        Sys.sleep(2)
        iris
      }, 
      server = FALSE, 
      options = list(lengthChange = FALSE)
    ) %>%
      bindCache(1)
    
    # Note that bindCache() must be used with renderDT(server=FALSE)
  }
)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,6 +37,8 @@ Suggests:
     rmarkdown,
     shiny (>= 1.2.0),
     testit
+Remotes:
+    rstudio/shiny
 VignetteBuilder: knitr
 RoxygenNote: 7.1.1
 Encoding: UTF-8
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -182,15 +182,12 @@ renderDataTable = function(
     )
   }
 
-  # This snapshotPreprocessOutput function was added in shiny 1.0.3.9002
-  if (exists("snapshotPreprocessOutput", asNamespace("shiny"))) {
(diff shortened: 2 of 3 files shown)


cpsievert (approved):
Just curious, is there a simple explanation why caching won't work with `server = T`?

wch:
Yeah, the reason it won't work is because when `server=T`, it sets state on the session, so that the client can send queries to it.

It calls `sessionDataURL()`, which calls `session$registerDataObj()`.

DESCRIPTION
```diff
@@ -37,6 +37,8 @@ Suggests:
     rmarkdown,
     shiny (>= 1.2.0),
     testit
+Remotes:
+    rstudio/shiny
```
yihui, line 41: Is there a plan for the next CRAN release of **shiny**? I'm wondering roughly when we could remove this `Remotes` field in **DT** (in weeks or months), so the release of **DT** won't be blocked. Thanks!

wch: We're planning to release in early January. The only reason that this Remotes entry is needed is because we link to the docs for a new function in Shiny. If you need to release earlier, we can remove that link and put it in a PR which we can merge later.

yihui: Got you. Thanks!

PR #885, opened 2020-12-07
Don't overwrite responsive details/options if provided

hdrab127:
Tiny change to allow passing responsive options like child row control and others: https://datatables.net/extensions/responsive/examples/child-rows/right-column.html 

Previously these would be overwritten by defaults when `options$responsive` set to `TRUE`.

Thanks heaps!
Hayden

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -256,7 +256,7 @@ datatable = function(
   params$extensions = if (length(extensions)) as.list(extensions)
 
   # automatically configure options and callback for extensions
-  if ('Responsive' %in% extensions & is.null(options$responsive)) {
+  if ('Responsive' %in% extensions && is.null(options$responsive)) {
     options$responsive = TRUE
   }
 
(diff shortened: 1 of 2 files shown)


hdrab127:
Note the responsive extension isn't compatible with crosstalk off the shelf. Will look into over the next few weeks.

R/datatables.R
```diff
@@ -256,7 +256,9 @@ datatable = function(
   params$extensions = if (length(extensions)) as.list(extensions)
 
   # automatically configure options and callback for extensions
-  if ('Responsive' %in% extensions) options$responsive = TRUE
+  if ('Responsive' %in% extensions & is.null(options$responsive)) {
```
shrektan, line 259:
```suggestion
  if ('Responsive' %in% extensions && is.null(options$responsive)) {
```

shrektan (approved):
Thanks. Would you mind add an item in NEWS.md?

yihui (approved):
Perfect. Thanks!

PR #907, opened 2021-04-08
`styleRow()` to style DT based on Row Indexes

### LINKED ISSUE #894: Format cell by position/row number rather than cell content
I hope I didn't miss anything, but I found no easy way to format a cell based on position rather than content 
(e.g change the background color of column "species", row 10 to red). 

The workarounds I found are kinda hackish, and *very* hackish and complicated to implement with editable DataTables.

Would it not be possible to implement something like a `rows` argument in format style? or `styleRows()`?

shrektan:
Closes #894 

```r
library(DT)
tbl = data.frame(
  COL_1 = c("A", "B", "C", "D"),
  COL_2 = c("E", "F", "G", "H"),
  stringsAsFactors = FALSE
)
datatable(tbl) %>% 
  formatStyle(
    columns = c(1),
    target = "cell",
    background = styleRow(list(2, c(1,3)), c("lightgreen", "pink"), default = "lightgrey")
  )
```

image


yihui (approved):
Looks great! Thanks!

PR #919, opened 2021-06-20
Implement vertical and opacity for noUiSlider filter

alexvpickering:
This adds a `vertical` and `opacity` option for filters and applies to the noUiSliders. I'm using this to implement DataTables with limited width that don't cause x overflow:

image

usage:

```R
datatable(iris, filter = list(position = 'top', vertical=TRUE, opacity=0.85))
```


yihui (approved):
Looks good to me! Please also add an item in `NEWS.md` and yourself as `ctb` to `DESCRIPTION`.

@shrektan Please feel free to merge if it looks good to you, too.

Thanks!

inst/htmlwidgets/datatables.js
```diff
@@ -499,9 +499,9 @@ HTMLWidgets.widget({
               // first, make sure the slider div leaves at least 20px between
               // the two (slider value) span's
               $x0.width(Math.max(160, $span1.outerWidth() + $span2.outerWidth() + 20));
-              // then, if the input is really wide, make the slider the same
-              // width as the input
-              if ($x0.outerWidth() < $input.outerWidth()) {
+              // then, if the input is really wide or slider is vertical,
+              // make the slider the same width as the input
+              if ($x0.outerWidth() < $input.outerWidth() | data.vertical) {
```
shrektan, line 504:
```suggestion
              if ($x0.outerWidth() < $input.outerWidth() || data.vertical) {
```

I think it should be `||`, correct?

shrektan (changes requested):
Looks good to me. Thanks.

inst/htmlwidgets/datatables.js
alexvpickering: That looks better to me as `|` is bitwise or

shrektan (approved):

shrektan:
Thank you for this PR!

Issue #934, opened 2021-09-28
filters do not refresh when data is modified


giadasp:
I have an editable datatable rendered in shiny.
I noticed that when a column has all missings or a single value, the filter on the top of the column is disabled (it is fine).
If I modify a value on a column with a disabled filter, let's say that I edit a c(1,1,1,1,1,1) column and I replace it with c(1,2,1,1,1,1) using `dataTableProxy `and` replaceData(resetPaging =FALSE, clearSelection="all")`, I tried also with `reloadData `e `clearSearch`, the column filter is still disabled.
It should happen instead that the datatable is refreshed and the filters get enabled.
Can you please provide a function or an argument to refresh the filters?

Also, I notice that the argument plain=TRUE, does not work. The numeric select input is still a slider.

Thank you
Giada

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('DT')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/DT')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

wholmes105:
@giadasp You can copy/paste `js_helper_functions` and `js_updateFilters` into your app's UI to use them; in any app with those inserted, `updateFilters()` will work fine as written.

@yihui I made a pull request, though I've never created a package before, so I may have made a rookie mistake when adding the code and documentation. Please let me know if anything needs fixing :)

yihui:
@wholmes105 I just merged the PR. Would you mind provide a simpler example to @giadasp with the dev version?

```r
remotes::install_github('rstudio/DT')
```

Thanks a lot!

wholmes105:
@yihui I am getting an error on my local machine that the function is unavailable, but I can see it in the updated code here on GitHub.
```Error: 'updateFilters' is not an exported object from 'namespace:DT'```

[...]

PR #939, opened 2021-10-29
Add updateFilters function and the associated JavaScript

wholmes105:
Adds an updateFilters() function that allows DT users to update the filters on their datatable, e.g. after updating the data in the table.


R/shiny.R
```diff
@@ -463,6 +463,36 @@ replaceData = function(proxy, data, ..., resetPaging = TRUE, clearSelection = 'a
...
+  # Calculate the values to be supplied to the filters based on column type
+  new_lims = lapply(data, function(x) {
+    if(inherits(x, c('numeric'))) {
+      range(x)
+    } else if(inherits(x, c('factor', 'logical'))) {
+      as.character(unique(x))
+    } else if(inherits(x, c('Date'))) {
+      as.numeric(as.POSIXct.Date(range(x))) * 100
+    } else if(inherits(x, 'POSIXt')) {
+      plyr::round_any(as.numeric(range(x)), accuracy = .01) * 100000
```
yihui, line 480: Is it possible to avoid introducing a new package dependency (i.e., plyr)?

R/shiny.R
```diff
@@ -463,6 +463,36 @@ replaceData = function(proxy, data, ..., resetPaging = TRUE, clearSelection = 'a
...
+    } else {
+      stop('updateFilters() requires all columns to be one of the following classes: numeric, factor, logical, Date, POSIXt')
+    }
+  })
+  
+  # As of DT 0.19, numeric values fed to the sliders need to be multiplied by 10; ex. 5.7 will be converted to 57
+  setNames(
```
yihui, line 487: Did you forget an assignment here?

```suggestion
  new_lims = setNames(
```

R/shiny.R
```diff
@@ -463,6 +463,36 @@ replaceData = function(proxy, data, ..., resetPaging = TRUE, clearSelection = 'a
...
 }
 
+#' @rdname proxy
+#' @rdname data
+#' 
+#' @export
+updateFilters = function(proxy, data) {
+  # Calculate the values to be supplied to the filters based on column type
+  new_lims = lapply(data, function(x) {
+    if(inherits(x, c('numeric'))) {
```
yihui, line 473: space after `if` (same thing for all other `if`s in both R and JS code):

```suggestion
    if (inherits(x, c('numeric'))) {
```

yihui (approved):
Sorry about the late response. The PR looks good overall. Just a few minor comments. It also needs a bullet in `NEWS.md`. You may add yourself as `ctb` in `DESCRIPTION`. Thanks!

yihui (approved):
Thanks a lot! I just made a few changes and hope you don't mind. I'll merge now. Please feel free to let me know if you want to make further changes.

yihui:
Sorry I didn't pay close enough attention to one implementation detail, and just fixed it. The current dev version should work now.

wholmes105:
I just downloaded the new dev version; everything worked as expected. Thanks!

bersbersbers:
I have the impression this PR breaks tables in `shinytest`, see #969

wholmes105:
@bersbersbers I'm not sure I know enough about `shinytest` to be of much help. Just to verify, have you confirmed that the bug did not exist as of DT 0.20.4 ?

bersbersbers:
@wholmes105 I think I have, see https://github.com/rstudio/DT/issues/969#issuecomment-1054118566

wholmes105:
@bersbersbers I see your list of commits, but I don't see dbfaf33 - could you check that? It's the commit for #961, which was meant to fix a bug with the tables not filling vertically. If the issue doesn't appear there, we can be sure it was introduced here

bersbersbers:
@wholmes105 I followed the order of commits in https://github.com/rstudio/DT/compare/v0.20...v0.21, so I did not check dbfaf33fad5568db1d5a1d9895b74245b9b27306 specifically because 55671f157469263874e84c9ebbbaa0b611efd585 (which came after dbfaf33fad5568db1d5a1d9895b74245b9b27306) was good, and 55671f157469263874e84c9ebbbaa0b611efd585 contains the fix. See
https://github.com/rstudio/DT/blob/55671f157469263874e84c9ebbbaa0b611efd585/inst/htmlwidgets/datatables.js#L1397

wholmes105:
@bersbersbers I was looking at the history of the DESCRIPTION file to see when it was updated with different versions; you are correct that 55671f1 is newer than dbfaf33, it just didn't appear in the list I was viewing because it was dbfaf33 that updated the DESCRIPTION file, not 55671f1. 

I agree with your comment in #969 that it is odd for simply adding a new function to cause this to break. Looking at the example you give, you don't even call the new function, so I am unsure what the cause could be.

bersbersbers:
> Looking at the example you give, you don't even call the new function, so I am unsure what the cause could be.

Indeed. I have checked the `shiny` and `shinytest` codes for strings like `"updateFilters"` (maybe these are some form of name collisions?), but did not find anything.

wholmes105:
Namespace collisions are about the only thing I can think of (without knowing much about `shinytest`) that might cause this.  The example I gave in #934 ran without issue in a shiny app, so we can probably narrow this down to just `shinyteest`. 

It seems unlikely that that `set_filter_lims` would be a collision, but perhaps the console in your web browser has some useful information? Does it return any errors or warnings when you run your app there instead of in RStudio?

bersbersbers:
> It seems unlikely that that `set_filter_lims` would be a collision, but perhaps the console in your web browser has some useful information? Does it return any errors or warnings when you run your app there instead of in RStudio?

Nothing at all, except a missing favicon. When I extend `test.R` like this
```r
app <- shinytest::ShinyDriver$new(".", debug="all", cleanLogs=FALSE)
app$enableDebugLogMessages(enable = TRUE)
app$snapshotInit("mytest")
app$snapshot()
print(app$getDebugLog("all"))
```
this is what I get:
```r
NULL
S/I 20:02:34.00 window.shinytest loaded
S/I 20:02:34.00 waiting for shiny session to connect
S/I 20:02:34.00 connected
S/I 20:02:34.00 ready
S/I 20:02:34.00 busy
S/I 20:02:34.00 idle
S/I 20:02:34.00 value tbl
S/I 20:02:34.00 message: {"values":[]}
C/E ----------- Loading required package: shiny
C/E -----------
C/E ----------- Attaching package: 'DT'
C/E -----------
C/E ----------- The following objects are masked from 'package:shiny':
C/E -----------
C/E -----------     dataTableOutput, renderDataTable
C/E -----------
C/E ----------- Running application in test mode.
C/E -----------
C/E ----------- Listening on http://127.0.0.1:5673
```

bersbersbers:
I tried to reproduce the issue with explicitly using `shinytest`, by isolating the effect of `shinytest::ShinyDriver$new` as seen in
https://github.com/rstudio/shinytest/blob/d0f6d737919b542c6a1e8bbd0ea05f9e4b3f5334/R/shiny-driver.R#L47
and
https://github.com/rstudio/shinytest/blob/d0f6d737919b542c6a1e8bbd0ea05f9e4b3f5334/R/initialize.R#L5
but without any luck.

Using the following as a launcher code launches the app at http://127.0.0.1:1234/, in test mode, and the table is rendered correctly. I wonder what else `shinytest` does.

```r
path = "."
shinyOptions = list(port = 1234)
is_rmd = \(x) FALSE
seed = 0
rng_kind = NULL
renderArgs = list()
options = list()
tempfile_format <- tempfile("%s-", fileext = ".log")

p <- withr::with_envvar(
	c("R_TESTS" = NA),
	callr::r_bg(
	  function(path, shinyOptions, rmd, seed, rng_kind, renderArgs, options) {

		if (!is.null(seed)) {
		  # Prior to R 3.6, RNGkind has 2 args, otherwise it has 3
		  do.call(RNGkind, as.list(rng_kind))
		  set.seed(seed);
		  getNamespace("shiny")$withPrivateSeed(set.seed(seed + 11))
		}

		options <- as.list(options)
		options$shiny.testmode <- TRUE
		do.call(base::options, options)

		if (rmd) {
		  # Shiny document
		  rmarkdown::run(path, shiny_args = shinyOptions, render_args = renderArgs)
		} else {
		  # Normal shiny app
		  do.call(shiny::runApp, c(path, shinyOptions))
		}
	  },
	  args = list(
		path = path,
		shinyOptions = shinyOptions,
		rmd = is_rmd(path),
		seed = seed,
		rng_kind = rng_kind,
		renderArgs = renderArgs,
		options = options
	  ),
	  stdout = sprintf(tempfile_format, "shiny-stdout"),
	  stderr = sprintf(tempfile_format, "shiny-stderr"),
	  supervise = TRUE
	)
)
```

bersbersbers:
I suspect it is an issue with PhantomJS after all.

`Rscript app.R` in one window, http://127.0.0.1:4271 looks fine in Chrome, but the following `test.js` run in `phantomjs test.js` yields the bottom `snap.png`:

```js
var page = require('webpage').create();
page.open('http://127.0.0.1:4271/', function() {
  page.render('snap.png');
  phantom.exit();
});
```

snap

And the reason the information is missing from the JSON as well is probably that these are generated by client-side javascript code.

No idea yet what the actual problem might be.

bersbersbers:
I think the problem is the use of the spread syntax, see https://github.com/ariya/phantomjs/issues/15187.

```js
var set_filter_lims = function(filter, new_vals) { var new_lims = [...new_vals]; };
```

When I paste the line above into the PhantomJS REPL, I get an error:

```
Unexpected token '.'

  phantomjs://repl-input:1 in global code
```

By contrast,
```
var set_filter_lims = function(filter, new_vals) { var new_lims = 1; };
```
works, so it's probably the `...` syntax.

Can you find a way of not using it? I have minimal JS experience, but it looks to me like `...` works like `*` in Python, and `[*vars]` can usually be replaced by `list(vars)`, so maybe there is another way to convert/clone `new_vals` into a new array,

yihui:
@wholmes105 How about using the `slice()` method to clone the array?

```diff
diff --git a/inst/htmlwidgets/datatables.js b/inst/htmlwidgets/datatables.js
index 27ed7ea..8ffb7bb 100644
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -95,7 +95,7 @@ var set_filter_lims = function(td, new_vals) {
 
   } else if (['number', 'integer', 'date', 'time'].includes(td.getAttribute('data-type'))) {
     // Note what the new limits will be just for this filter
-    var new_lims = [...new_vals];
+    var new_lims = new_vals.slice(); 
     // Determine the current values and limits
     var slider = $(td).find('.noUi-target').eq(0);
```

wholmes105:
PR #970 implements the change to patch this

PR #947, opened 2021-11-30
Documentation for editData changed for parameter info

gsmolinski:
This PR adresses #941 issue.

For parameter `info` (in `editData()` function) documentation indicated that returning object is a `data.frame` with columns: `row`, `column`, `value`, but second name should be `col`. This is problematic for the user, because when she / he refers to `object$column` (instead of `object$col`), `NULL` is returned.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -109,7 +109,7 @@ coerceValue = function(val, old) {
 #' @param data The original data object used in the DataTable.
 #' @param info The information about the edited cells. It should be obtained
 #'   from \code{input$tableId_cell_edit} from Shiny, and is a data frame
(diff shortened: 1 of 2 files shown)


shrektan:
Can you please build the package so that the rd file can be updated as well?

gsmolinski:
Sorry. Now I have used `document()`, `.Rd` file should be up to date. However, my experience is very limited, especially when talking about git, so I don't know if I did it correctly. `document()` function changed more than just one file, but I decided to commit and push only the file related to `editData()` function. If I'm doing something wrong, I can try one more time, but if you think that it will be better for you to do this on your own, you can just close this PR

yihui (approved):
Looks good now. Thank you very much!

PR #512, opened 2018-02-25
classNameDefinedColumns() now identifies negative and "_all" targets

shrektan:
closes #511 

The fix #476 misses the fact that besides of column indexes, there're [4 kinds](https://datatables.net/reference/option/columnDefs.targets) of `columnDefs.targets` in total.

- [x] 0 or a positive integer - column index counting from the left
- [x]  A negative integer - column index counting from the right
- [ ] A string - class name will be matched on the TH for the column (without a leading .)
- [x] The string "_all" - all columns (i.e. assign a default)

This PR is meant to handle all the rest cases. **However, for case 3, I can't reproduce the effect in the first place.** What I mean is explained by the following example:

```r
DT::datatable(
  data.frame(a = 1, b = 2, c = 3, d = 4),
  options = list(
    columnDefs = list(
      list(className = "dt_center", targets = "a") 
      # it should have set "a" column center alignment, but it didn't
    )
  )
)
```

For 1, 2 and 4 cases, they are correctly handled now.

```r
DT::datatable(
  data.frame(a = 1, b = 2, c = 3, d = 4),
  options = list(
    columnDefs = list(
      list(className = "dt-center", targets = "_all")
    )
  )
)

DT::datatable(
  data.frame(a = 1, b = 2, c = 3, d = 4),
  options = list(
    columnDefs = list(
      list(className = "dt-center", targets = c(-1, 2))
    )
  )
)

```

PR #948, opened 2021-12-04
Allow string options$columnDefs$targets

shrektan:
Due to implentation of DT, `options$columnDefs$targets` doesn't support string values (See #512). 

It's inconvinient as using column-names is always easier. In addition, it confuses the user in the first place, as it is inconsistent with DataTables' documentation.

This PR addresses this issue by converting the string valued `targets` into column index values of the data, internally.

### CODE

```r
x <- rnorm(10)
rk <- rank(x)
prefix <- sample(LETTERS[1:3], replace = TRUE, 10)
x_txt <- sprintf("%s%0.2f%%", prefix, x * 100)
tbl <- data.frame(V1 = x, V2 = x_txt, V3 = x, RK = rk)
DT::datatable(tbl, options = list(
  scrollY=FALSE, 
  columnDefs = list(
    list(visible = FALSE, targets = list("V3", 2))
  )
))
```

### Before the PR ("V3" is visible but it should not)

image

### After the PR ( "V3" is now invisible, as expected)

image

### DIFF
--- a/tests/testit/test-datatables.R
+++ b/tests/testit/test-datatables.R
@@ -196,8 +196,9 @@ assert('warn autoHideNavigation if no pageLength', {
 assert("colDefsTgtHandle() works", {
   cols = c("A", "B", "C")
   (colDefsTgtHandle(NULL, cols) %==% list())
-  (colDefsTgtHandle(NULL, cols) %==% list())
-
+  (colDefsTgtHandle(list(), cols) %==% list())
+  (has_error(colDefsTgtHandle("abc", cols)))
+  (has_error(colDefsTgtHandle(list("abc"), cols)))
   defs = list(
     list(1, targets = "_all"),
     list(2, targets = 1L),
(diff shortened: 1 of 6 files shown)


yihui (approved):
Sounds great! Feel free to merge the PR by yourself whenever you are ready (I only had a quick look). Thanks!

tests/testit/test-datatables.R
```diff
@@ -192,3 +192,31 @@ assert('warn autoHideNavigation if no pageLength', {
...
+    list(5, targets = list(c("A", "C"), "_all")),
+    list(6, targets = list(1L, "_all")),
+    list(7, targets = list(1L, "C")),
+    list(8, targets = list(1L, "B", "_all"))
+  )
+  res = list(
+    list(1, targets = "_all"),
+    list(2, targets = 1L),
+    list(3, targets = 1L),
+    list(4, targets = list("_all", 0L)),
```
shrektan, line 215: I can't remember why it should return `list("_all", 0L)` instead of `list("_all")`. What's the difference??

PR #949, opened 2021-12-05
Enabling R styling via a new param `formatter`

### LINKED ISSUE #938: Scientific notation format function for DT
Hi,
I'm trying to format some columns with scientific notation and finding it very difficult. I have been trying to write a short javascript snippet to take advantage of toExponential using rowCallback, but to no avail (see example in [https://stackoverflow.com/questions/49634329/how-to-force-scientific-notation-with-renderdt](https://stackoverflow.com/questions/49634329/how-to-force-scientific-notation-with-renderdt)).

I was wondering if a formatScientific function could be created, as the ones that DT already has (formatCurrency, formatRound, etc ...)

This formatScientific could use as argument the number of decimal digits to use for the significand (or mantissa) in order to get a number with normalized scientific notation [(https://en.wikipedia.org/wiki/Scientific_notation)](https://en.wikipedia.org/wiki/Scientific_notation)

Thank you very much to all the contributors in this repo. Your work is highly appreciated!

---

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

shrektan:
Closes #938 #652 

DT implements many nice JS wrappers on styling the columns. However, R users may still find they want to implement their own style, as the love for beautiful table never ends.  They can do it in two ways:

1. On JS side: Write a JS function and set it in `render` options, i.e., `datatable(..., options = list( columnDefs = list( list(targets = 1, render = JS(...)))))`
2. On R side: Transform the R data.frame object's relevent column into characters, then pass it to DT, probably need to set the `escape` argument to `FALSE`.

Unfortunately, those two methods are either difficult to implement or contain side-effects:

1. JS method: 
  - Many R users are not aware of the existence of the `columnDefs.render` option
  - Write the JS function correctly is not easy, especially when the "style" is complicated
  - You can't leverage the existing R packages like `formattable`
2. R method:
  - Probably lose the ability to sort, filter or search correctly as the raw "data" is tranformed into formatted "data"
  - Probably need additional efforts to "unescape" data, align text, etc.

The new param `formatter` allows users to write formatting functions in R without losing the ability to sort, filter or search by  retaining the raw data "untouched" by creating invisible helper columns.

## NOTE

This PR depends on PR #948 

## TODO

- [ ] support formatter type like `list(function() ... , targets = ...)`
- [ ] figure out a way when the formatter need other info from the data, like other column
- [ ] try out all the possible corner cases in case of unexpected happening
- [ ] state clear the limitations in doc (probably cause issues when editing data
- [ ] need to triger updating the hidden formatted columns when source is updated, e.g, via proxy, or editing
- [ ] NEWS 
- [ ] Add an example into inst/example
- [ ] Add an example into site
- [ ] add tests

## Example

```r
set.seed(100)
x <- rnorm(10)
rk <- rank(x)
prefix <- sample(LETTERS[1:3], replace = TRUE, 10)
tbl <- data.frame(V1 = x, V2 = x, V3 = x, V4 = x, RK = rk)
DT::datatable(tbl, rownames = FALSE, filter = "top", formatter = list(
  V2 = function(x) paste0(prefix, round(x * 100, 2), "%"),
  V3 = formattable::percent,
  V4 = function(x) round(x * 100, 2),
  V4 = formattable::color_tile(min.color = "yellow", max.color = "red")
))
```
image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.20.1
+Version: 0.20.2
 Authors@R: c(
     person("Yihui", "Xie", email = "xie@yihui.name", role = c("aut", "cre")),
     person("Joe", "Cheng", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # CHANGES IN DT VERSION 0.21
 
+## MAJOR CHANGES
+
+- Now users can provide column names of the data to `options$columnDefs$targets`. Previously, it only supports column indexes or "_all" (thanks, @shrektan #948).
 
 # CHANGES IN DT VERSION 0.20
 
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -262,6 +262,20 @@ datatable = function(
   # disable CSS classes for ordered columns
   if (is.null(options[['orderClasses']])) options$orderClasses = FALSE
 
+  data = applyFormatter(data, formatter)
+  format_cols = attr(data, "DT.format.cols", exact = TRUE)
+  raw_cols = attr(data, "DT.raw.cols", exact = TRUE)
+  if (length(format_cols)) options = appendColumnDefs(options, list(
+    visible = FALSE, targets = targetIdx(format_cols, base::colnames(data), showRowName = !is.null(rn))
+  ))
+  options$columnDefs = append(
+    options$columnDefs, colFormatter(
+      raw_cols, base::colnames(data), rownames = !is.null(rn), template = function(format_col_idx) {
+        sprintf("row[%d];", format_col_idx)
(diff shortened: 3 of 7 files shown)


shrektan:
Now I think this feature is not suitable for DT as it becomes more complicated thant I thought. I don't wanna introduce any complexity into DT as it's already a big project. 

Close it for now.

philibe:
It's was a vert good try. Thanks.  (Sob). Perhaps later ? :)

philibe:
@yihui , @shrektan

This reopened issue gives me the opportunity to thank you for your work on DT.

When I first installed DT in early 2018, for me it was a simple adaptation of datatable.net, but now that I'm trying to use datatable.net with Flask Rest and VueJs 3 I see that I have to rewrite a lot of things  which I thought was in datatable.net but actually written in the server side of DT, and I understand how your API wrapping initial datatable.net ones.

PR #956, opened 2022-01-24
repeat the scalar values argument in `styleRow()` and `styleEqual()`

### LINKED ISSUE #955: Repeat scalar value for styleEqual and styleRow
Currently, the following codes result in errors that " length(levels) must be equal to length(values)". I thinks we'd better repeat the "values" as it's scalar. This fits the convention of R.

```r
DT::styleEqual(LETTERS, "black")
DT::styleRow(1:3, "red")
```

shrektan:
Closes #955 

```r
DT::datatable(
  data.frame(A = letters, B = LETTERS)
) |> 
  DT::formatStyle("A", backgroundColor = styleRow(2:4, "pink")) |>
  DT::formatStyle("B", backgroundColor = styleEqual(c("C", "D"), "lightblue"))
```

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: DT
 Type: Package
 Title: A Wrapper of the JavaScript Library 'DataTables'
-Version: 0.20.2
+Version: 0.20.3
 Authors@R: c(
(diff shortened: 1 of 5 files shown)


yihui (approved):
Thanks!

PR #962, opened 2022-02-01
Be able to specify the styled rows in `formatXXX()`

### LINKED ISSUE #520: formatCurrency() on some rows only
formatCurrency() applies to an entire column. How can I apply this type of formatting to only some rows?

styleInterval() and styleEqual() seems to work only with formatStyle().

shrektan:
Closes #520

# Code

```r
library(shiny)
library(DT)

data <- iris
data$DATE <- sample(0:500, replace = TRUE, size = nrow(data)) + as.Date("1990-01-01")
rowsOpt = list("top3and12" = c(1:3, 12), "1/5/12/20" = c(1,5,12,20), "all" = NULL)

ui = fluidPage(
  radioButtons("rows", "rows", names(rowsOpt), inline = TRUE),
  radioButtons("fmtfun", "format function", c(
    "formatCurrency", "formatDate", "formatPercentage", "formatRound", "formatSignif", "formatString"
  ), inline = TRUE),
  selectInput("cols", "columns", seq_len(ncol(data)), selected = c(1,3,6), multiple = TRUE),
  DTOutput("tbl")
)
server = function(input, output, session) {
  output$tbl = renderDT({
    styler = function(data) {
      args = list(
        columns = as.integer(input$cols |> req()), rows = rowsOpt[[input$rows]]
      )
      # formatString's default arg won't change anything...
      if (input$fmtfun == "formatString") {
        args = c(args, list(suffix = "--$$"))
      }
      args = c(list(data), args)
      fun = match.fun(input$fmtfun)
      do.call(fun, args)
    }
    datatable(data) |>
      styler()
  }, server = FALSE)
}
runApp(mget(c("ui", "server")))
```

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 - Add the `zero.print` argument to `formatPercentage()`, `formatCurrency()`, `formatSignif()` and `formatRound()`, which allows to control the format of zero values. It's useful when the data is "sparse" (thanks, @shrektan #953).
 
-- `formatXXX()` functions now gain a new argument `rows` (starting from 1), which can be used to specify the rows that the "style" should be applied to. Note that it only works expected in the client-side processing mode, i.e., `server = FALSE`. `formatStyle()` is the only exception that doesn't have this argument and should use `styleRow()` instead. (thanks, @jrecasens @shrektan #520)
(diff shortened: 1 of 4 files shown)


yihui (approved):
Looks good. Thanks!

Issue #969, opened 2022-02-28
Tables don't render in `shinytest` since 0.21


bersbersbers:
`app.R`
```r
library(shiny)
library(DT)
shinyApp(
  ui = fluidPage("Hello", DTOutput("tbl"), "World"),
  server = function(input, output) {
    output$tbl <- renderDT(
      iris,
      options = list(lengthChange = FALSE)
    )
  }
)
```

`test.R`
```r
app <- shinytest::ShinyDriver$new(".")
app$snapshotInit("test")
app$snapshot()
```

Now run:
```
mkdir tests
Rscript -e "devtools::install_version('DT', version='0.20')"
Rscript test.R
```

`tests\test-current\001.png`:
![001](https://user-images.githubusercontent.com/12128514/155965067-6efbad4e-2b78-4b0a-bf72-3614b25ecac7.png)
(The rendering is not ideal, but you do see the table in some way.)

Then run:
```
Rscript -e "devtools::install_version('DT', version='0.21')"
Rscript test.R
```

`tests\test-current\001.png`:
![001](https://user-images.githubusercontent.com/12128514/155965442-a2914f28-5629-4918-a661-826889308360.png)

Finally run:
```
Rscript -e "remotes::install_github('rstudio/DT')"
Rscript test.R
```

`tests\test-current\001.png`:
(the same as with 0.21 above)

Note that the table is missing from `shinytest`s `.json` snapshots as well, so this is somewhat unlikely to be a pure client-side rendering issue.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('DT')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/DT')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

I can repro the issue on Windows and on Linux.
```
R version 4.1.2 (2021-11-01)
Platform: x86_64-w64-mingw32/x64 (64-bit)
[...]

PR #970, opened 2022-03-01
Patch compatibility with shinytest and PhantomJS

wholmes105:
PhantomJS had issues with the spread operator `...` used in `set_filter_lims`, which caused datatables to not appear when DT was combined with the shinytest package. `...` has been replaced with an equivalent operation using the `slice` method.


bersbersbers:
If successfull, this should close #969.

yihui (approved):
Thanks!

bersbersbers:
Awesome, thanks - I'll try this tomorrow!

Issue #971, opened 2022-03-09
`updateFilters()` enhancements


mikmart:
It was great to see the new `updateFilters()` function added recently – thanks @wholmes105!

After some testing, there’s a couple of useful enhancements I had in mind:

1.  Currently `updateFilters()` errors out if the data contains integer or character columns. I don’t think there’s any reason for this behaviour? Integer filters should update like other numerics, and character can just be skipped.
2.  It would be nice to have the updated limits match those that would be calculated if the table were re-rendered with the new data. I believe this would require refactoring the internal `filterRow()` function so as to be able to re-use the limit calculations there.
    -   For example factors should use `levels()`, not `unique()` for limits.
3.  Limits for numeric columns currently ignore the internal `scale` applied to the slider widgets. As a result, the updated limits are correct only when the column happens to have a maximum of 1 decimal places in the data (or 0 for dates, 3 for datetimes).
4.  Filters that were `disabled` on the initial render don’t have their limits updated or enabled. That’s because the HTML for them is never created. Fixing this might require a more thorough refactoring for `filterRow()`.

Here’s a reprex app demonstrating some of the issues I mentioned:

``` r
library(shiny)
library(DT)

tbl <- data.frame(
  num1 = seq(-2, 2) / 1,
  num2 = seq(-2, 2) / 10,
  num3 = seq(-2, 2) / 100,
  num4 = replace(seq(-2, 2) / 100, 1, Inf),
  dttm = round(Sys.time()) + seq(-2, 2) * 3600,
  fct1 = factor(c("A", rep("B", 4)), levels = c("A", "B", "C")),
  fct2 = factor(c(rep("A", 4), "B"), levels = c("A", "B", "C"))
)

ui <- fluidPage(
  sliderInput(
    "rows",
    "Slice rows",
    min = 0,
    max = nrow(tbl),
    value = c(1, 4),
    step = 1
  ),
  DTOutput("table"),
)

server <- function(input, output, session) {
  output$table <- renderDT(datatable(tbl, filter = "top"))
  
  tbl_slice <- eventReactive(input$rows, {
    r1 <- input$rows[1]
    r2 <- input$rows[2]
    tbl[seq(r1, r2), ]
  })
  
  proxy <- dataTableProxy("table")
  observeEvent(tbl_slice(), {
    replaceData(proxy, tbl_slice())
    updateFilters(proxy, tbl_slice())
  })
}

shinyApp(ui, server)
```

[...]

PR #973, opened 2022-03-09
Fix updating filters with length 1 limits

mikmart:
Part of #971.

If `updateFilters()` sets tries to set new limits for a select filter that would have length 1, the data in the update message gets unboxed into a scalar. However, the JS code that accesses these values expects the message to always contain only arrays.


yihui (approved):
Perfect. Thanks!

PR #974, opened 2022-03-09
Use filter scale when updating limits

mikmart:
Part of #971.

Updated numeric limits were only correct for reals with 1 significant digit, dates with 0 significant digits, and datetimes with 3 significant digits. That's because they didn't take into account the `scale` applied to the internal representation of the slider limits.

This PR updates `updateLimits()` to rescale the new limits before applying them, fixing the issue.


yihui (approved):
Great! Thanks!

PR #975, opened 2022-03-09
Update factors and logicals consistently with initial render

mikmart:
Part of #971.

`updateFilter()` should apply new limits consistently with how they would appear if fully re-rendered (and thus calculated internally with `filterRow()`). Currently booleans always show all options, and `NA` if there are any, while factors let you pick from their levels, regardless of whether or not they are present in the data:

https://github.com/rstudio/DT/blob/815872d67141bb053ed9defd6b72cbaed2aa6668/R/datatables.R#L684-L692

This PR changes the new limits for factors and booleans calculated by `updateFilter()` to match those found by `filterRow()`.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -479,8 +479,10 @@ updateFilters = function(proxy, data) {
   new_lims = lapply(data, function(x) {
     if (inherits(x, c('numeric'))) {
       range(x)
-    } else if (inherits(x, c('factor', 'logical'))) {
-      as.character(unique(x))
+    } else if (is.logical(x)) {
+      c("true", "false", if (anyNA(x)) "na")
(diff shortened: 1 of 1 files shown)


mikmart:
In general, it might be nicer to refactor `filteRow()` so that the part that calulates the new limits could be re-used in `updateFilter()` to ensure consistency. However, I thought I'd leave that for later, in case there's interest in that.

yihui (approved):
> In general, it might be nicer to refactor `filteRow()` so that the part that calulates the new limits could be re-used in `updateFilter()` to ensure consistency.

Absolutely! Please feel free to do that after your PRs are merged. Thanks!

mikmart:
Great, I'll have a look at that then!

PR #977, opened 2022-03-11
Refactor `filterRow()` and update disabled state in `updateFilters()`

mikmart:
Part of #971.

This PR factors out the filter property calculations (like limits and disabled status) from `filterRow()`  into a new `columnFilters()` method. That common method is then used for both the initial render, and subsequent updates with `updateFilters()`, ensuring consistency.

As it flowed naturally into the above changes, I also updated the JS side for `updateFilters()` to also manage the enabled/disabled state of the controls.

There's a bit more going on in this PR, but I tried to group my commits sensibly. Hopefully that helps with review.

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -625,13 +625,25 @@ tableHead = function(names, type = c('head', 'foot'), escape = TRUE, ...) {
   f(tags$tr(lapply(escapeColNames(names, escape), tags$th)), ...)
 }
 
-#' @importFrom htmltools tagList
 filterRow = function(
   data, rownames = TRUE,
   filter = list(position = 'none', clear = TRUE, plain = FALSE, vertical = FALSE, opacity = 1)
 ) {
   if (filter$position == 'none') return()
-  tds = list()
+
+  filters = columnFilters(data)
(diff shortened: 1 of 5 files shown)


R/datatables.R
```diff
@@ -625,90 +625,141 @@ tableHead = function(names, type = c('head', 'foot'), escape = TRUE, ...) {
...
+    ctrl = if (f$control == 'slider') {
+      tags$div(
+        style = paste0('display: none;position: absolute;width: 200px;opacity: ', options$opacity),
+        tags$div(`data-min` = p$min, `data-max` = p$max, `data-scale` = p$scale),
+        if (options$vertical) tagList(
+          tags$span(style = 'position: absolute; bottom: 0px; left: 15px;'),
+          tags$span(style = 'display: none;', HTML('&nbsp;')),
+          tags$span(style = 'position: absolute; top: 2px; left: 15px;')
+        )
+        else tagList(
```
yihui, line 738:
```suggestion
        ) else tagList(
```

yihui (approved):
Looks good! Thank you!

PR #978, opened 2022-03-11
Fix `updateFilters()` with un-searchable columns

### LINKED ISSUE #971: `updateFilters()` enhancements

mikmart:
Closes #971.

- Fixed an issue where updating filters was writing to instead of reading the `searchable` status of the filter when determining whether or not they should be disabled. This was introduced in #977 so hasn't made it into "newsworthy" status yet.
- Added an example of missing values in sliders to the demo app.
- Some minor stylistic changes to better fit the code-base style.

### DIFF
--- a/inst/examples/DT-updateFilters/app.R
+++ b/inst/examples/DT-updateFilters/app.R
@@ -2,10 +2,10 @@ library(shiny)
 library(DT)
 
 tbl <- data.frame(
-  int = -2:2,
+  int = seq(-2, 2),
   num1 = seq(-2, 2) / 10,
   num2 = replace(seq(-2, 2) * 10, 1, -Inf),
-  date = Sys.Date() + seq(-2, 2),
+  date = replace(Sys.Date() + seq(-2, 2), 1, NA),
   dttm = round(Sys.time()) + seq(-2, 2) * 3600,
   fct1 = factor(c("A", rep("B", 3), "C")),
   fct2 = factor(c(rep("A", 4), "B"), levels = c("A", "B", "C")),
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -74,7 +74,7 @@ window.DTWidget = DTWidget;
 // A helper function to update the properties of existing filters
 var setFilterProps = function(td, props) {
   // Update enabled/disabled state
-  var $input = $(td).find("input").first();
+  var $input = $(td).find('input').first();
   var searchable = $input.data('searchable');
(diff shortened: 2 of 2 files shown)


yihui (approved):
Great. Thanks!

Issue #587, opened 2018-09-05
Column filters not updating


dcleere:
Hi there,

My issue is identical to what was reported in StackOverflow here - https://stackoverflow.com/questions/45547670/datatable-filters-update-on-the-fly

When I filter by one column, the factors in the remaining columns include all possibilities, regardless of whether or not it is relevant to the first filter.  Can this be updated?

Thanks!

rfineman:
Any progress on this front? Facing the same issue.

shrektan:
Sorry for the late reply but there's no short plan on this feature yet. However, there's a new extension [SearchPanel](https://datatables.net/blog/2020-01-13) that is able to do that (but it's different from the `filter` and only works for the client-side processing mode).  This new extension will be available when #756 gets merged.

AhmedKhaled945:
@shrektan Hello, is this still the case?

yihui:
@AhmedKhaled945 I think it should be possible now with @wholmes105 and @mikmart's recent work.

```r
remotes::install_github('rstudio/DT')
```

Mikko has provided an example here: https://github.com/rstudio/DT/blob/ddcc9e91ffa00924ccffeb6ceee14dfb2e14c0b6/inst/examples/DT-updateFilters/app.R#L46

You can update the filters with the filtered data, which you can obtain from `input$tableId_rows_all` (https://rstudio.github.io/DT/shiny.html). It should work for both client-side and server-side processing.

That said, the SearchPanes extension may also work, but it would only work for the client-side processing mode as @shrektan mentioned above.

AhmedKhaled945:
This is the example found, tried to tweak it to implement dependent filters, i think there is a way but somehow it is not working as expected, it gives wrong factor levels on the last 2 columns and it revert filtering when i filter using the sliders,

```r 
library(shiny)
library(DT)

tbl <- data.frame(
    num1 = seq(-2, 2) / 1,
    num2 = seq(-2, 2) / 10,
    num3 = seq(-2, 2) / 100,
    num4 = replace(seq(-2, 2) / 100, 1, Inf),
    dttm = round(Sys.time()) + seq(-2, 2) * 3600,
    fct1 = factor(c("A", rep("B", 4)), levels = c("A", "B", "C")),
    fct2 = factor(c(rep("A", 4), "B"), levels = c("A", "B", "C"))
)

ui <- fluidPage(
    sliderInput(
        "rows",
        "Slice rows",
        min = 0,
        max = nrow(tbl),
[...]

PR #982, opened 2022-03-17
Extract and export server-side searching functions

mikmart:
The linked filter feature requested in #587 would benefit from being able to redo server-side searches.

This PR extracts and exports the global and column searching parts from the internal `dataTablesFilter()` function as `doGlobalSearch()` and `doColumnSearch()` respectively, letting users run modified searches. This is useful in conjunction with the new `updateFilters()` for setting filter limits based on the values that would be available if all _other_ filters were applied.

Here's an example app demonstrating the kind of functionality enabled by this feature:

``` r
library(shiny)
library(DT) # remotes::install_github("mikmart/DT@export-search")
library(dplyr)

# Adverse events data from a fake clinical trial
file <- file.path(tempdir(), "adae.xpt")
if (!file.exists(file)) {
  url <- "https://github.com/RConsortium/submissions-pilot1-to-fda/raw/main/m5/datasets/rconsortiumpilot1/analysis/adam/datasets/adae.xpt"
  curl::curl_download(url, file, quiet = FALSE)
}
adae <- haven::read_xpt(file)

tbl <- adae %>% 
  transmute(
    USUBJID = factor(USUBJID),
    AESEQ = as.integer(AESEQ),
    AEDECOD = factor(AEDECOD),
    AESOC = factor(AESOC)
  )

ui <- fluidPage(
  DTOutput("table")
)

  output$table <- renderDT(tbl, filter = "top")
  
  filterable_sets <- eventReactive(input$table_search_columns, {
    # Get separately filtered indices
    fi <- Map(doColumnSearch, tbl, input$table_search_columns)
    
    # Find what rows others leave available
    ai <- lapply(seq_along(fi), function(j) Reduce(intersect, fi[-j]))
    
    # Get the corresponding data
    lapply(Map(`[`, tbl, ai), function(x) {
      if (is.factor(x)) droplevels(x) else x
    })
  })
  
  observeEvent(filterable_sets(), {
    updateFilters(proxy, filterable_sets())
  })
}

shinyApp(ui, server)
```


yihui (approved):
From the experience of the past PRs, I guess I can just trust your PRs now, since you've been very meticulous. I didn't review this PR on a line-by-line basis, but it looks good to me overall. If I fail to recognize any potential problems, we can definitely amend this PR later. Thanks!

mikmart:
Thank you for the trust @yihui!

mikmart:
@philippleppert you can use `doGlobalSearch()` to find the rows that the global search box filters out, and also exclude those values from each filterable set. Modifying the example here, you'd add something like:

``` r
    # Account for global search
    gi <- doGlobalSearch(as.data.frame(tbl), input$table_search)
    ai <- lapply(ai, intersect, gi)
```

(There's a bug in `doGlobalSearch()` at the moment which makes it not work with tibbles, so need `as.data.frame()`.)

You also need to make sure `filterable_sets()` takes a reactive dependency on the search box input:

``` r
  filterable_sets <- eventReactive(c(input$table_search, input$table_seach_columns), {
```

philippleppert:
@mikmart this works fine! thank you so much 👍

PR #984, opened 2022-03-19
Server-side search optimizations

mikmart:
While working on #982 I thought of two optimizations to speed up server-side searching. This PR:

1. Doesn't bother searching rows that have already been ruled out by earlier searches.
2. Does the (much more expensive) global search after the (quicker) column searches have already been applied.

I also added some infrastructure to help test the server-side search handler `dataTablesFilter()`.

Benchmarking, I saw a ~20x speed improvement for a 340k x 6 dataset with a global search and two fairly specific column filters:

``` r
library(nycflights13)

# Keep the client query a reasonable size with a subset of columns
flights0 <- flights[c("year", "month", "day", "carrier", "flight", "tailnum")]
flights0 <- as.data.frame(flights0)

# Client-side query with global search and two column filters
query = list(
  draw = "0", start = "0", length = "10", escape = "true",
  search = list(
    value = "N6", regex = "false", caseInsensitive = "true",
    smart = "true"
  ), columns = list(
    `0` = list(
      searchable = "true",
      search = list(value = "", regex = "false")
    ), `1` = list(
      searchable = "true", search = list(
        value = "1 ... 1",
        regex = "false"
      )
    ), `2` = list(
      searchable = "true",
      search = list(value = "", regex = "false")
    ), `3` = list(
      searchable = "true", search = list(value = "UA", regex = "false")
    ),
    `4` = list(searchable = "true", search = list(
      value = "",
      regex = "false"
    )), `5` = list(
      searchable = "true",
      search = list(value = "", regex = "false")
    )
  )
)

# Before this branch
system.time(res0 <- DT:::dataTablesFilter(flights0, query))
#>    user  system elapsed 
#>    0.5     0.0     0.5 

# With this branch
devtools::load_all()
system.time(res1 <- dataTablesFilter(flights0, query))
#>    user  system elapsed 
#>    0.00    0.00    0.02 

identical(res0, res1)
#> [1] TRUE
```

### DIFF
--- a/tests/testit/test-search.R
+++ b/tests/testit/test-search.R
@@ -1,5 +1,9 @@
 library(testit)
 
+# Factors and strings are searched differently.
+# Older versions of R don't have this set.
+op <- options(stringsAsFactors = FALSE)
+
 assert('searching integer values works', {
   x = seq(-2, 2)
   (setequal(doColumnSearch(x, '0 ... 2'), 3:5))
@@ -159,3 +163,6 @@ assert('server-side search handler ignores NULL search', {
   out = dataTablesFilter(tbl, query)
   (setequal(out$DT_rows_all, 1:3))
 })
+
+# Restore stringsAsFactors
+options(op)
--- a/tests/testit/test-search.R
+++ b/tests/testit/test-search.R
@@ -2,7 +2,7 @@ library(testit)
 
 # Factors and strings are searched differently.
 # Older versions of R don't have this set.
-op <- options(stringsAsFactors = FALSE)
+op = options(stringsAsFactors = FALSE)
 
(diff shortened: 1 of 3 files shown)


tests/testit/test-search.R
```diff
@@ -1,5 +1,9 @@
 library(testit)
 
+# Factors and strings are searched differently.
+# Older versions of R don't have this set.
+op <- options(stringsAsFactors = FALSE)
```
yihui, line 5:
```suggestion
op = options(stringsAsFactors = FALSE)
```

yihui (approved):
Awesome! I wished to optimize this quite a few years ago, and I'm really glad that someone finally did it. Thanks!

PR #1017, opened 2022-09-09
update to jszip 3.10.1

hedsnz:
Update jszip to 3.10.1 to address #1016.

Source of jszip: https://github.com/Stuk/jszip/blob/main/dist/jszip.min.js

I'm not really clear on how jszip is called internally, so I haven't written any sophisticated tests, just verifying that the buttons are rendered correctly. And just verified manually that the buttons are working correctly by downloading the .xlsx file as per below:

```
datatable(iris, extensions = "Buttons", options = list(
    dom = "Bt", buttons = c("excel", "csv", "pdf"))
)
```


tests/testit/test-datatables.R
```diff
@@ -223,3 +223,10 @@ assert("colDefsTgtHandle() works", {
   )
   (colDefsTgtHandle(defs, cols) %==% res)
 })
+
+assert("buttons load", {
+  out <- datatable(iris, extensions = "Buttons", options = list(
+    dom = "Bt", buttons = c("excel", "csv", "pdf")))
+  out$x$extensions[[1]][1] %==% "Buttons"
+  unlist(out$x$options$buttons) %==% c("excel", "csv", "pdf")
```
yihui, line 231:
```suggestion
  (out$x$extensions[[1]][1] %==% "Buttons")
  (unlist(out$x$options$buttons) %==% c("excel", "csv", "pdf"))
```

yihui (approved):
Sounds good. Thank you!

I'll merge this PR for now. This change really needs to be done in the upstream DataTables library. We are not using the latest version of DataTables yet. I wonder if they have updated jszip there.

PR #1020, opened 2022-09-19
two new examples

stla:
Added two examples which use a JavaScript function in the options.

### DIFF
--- a/inst/examples/DT-searchExact/server.R
+++ b/inst/examples/DT-searchExact/server.R
@@ -3,12 +3,12 @@ library(DT)
 
 function(input, output, session) {
 
-  dat <- data.frame(
+  dat = data.frame(
     car = c("Mazda", "Mazda RX4", "Mazda RX4 Wag", "Ford", "Mercedes"),
     pet = c("dog", "dog", "cat", "cat", "cat")
   )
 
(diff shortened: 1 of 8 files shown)


yihui (approved):
Thanks!

inst/examples/DT-searchExact/server.R
```diff
@@ -0,0 +1,56 @@
...
+      "  $inputs.off('keyup search input').on('keyup', function() {",
+      "    var value = $(this).val();",
+      "    if(value !== '') {",
+      "      value = '^' + value + '$';",
+      "      var index = 1 + $inputs.index(this);", # add one for rownames column
+      "      var column = table.column(index);",
+      "      column.search(value, true, false).draw();",
+      "    }",
+      "  });",
+      "}"
```
yihui, line 41: For large amounts of JS code, I wonder if it'd better to write them in a separate JS file and read the lines here. That might make the JS code easier to read and develop.

stla: But is it possible to make this JS file visible with the `showcase` option of Shiny?

yihui: Good point. No.

Okay, I'll just merge it as is. Thanks!

PR #1022, opened 2022-09-27
Default to `width:100%;height:auto` when statically rendering in browser

### LINKED ISSUE #1021: Static render of datatable() defaults to fixed size of 500 x 960
For example, if you open the result of `DT::datatable(mtcars)` in a browser window, you get that arbitrary fixed size. It seems the default size in this case should match the shiny & knitr defaults of `height: auto; width: 100%`

cpsievert:
Closes #1021


yihui (approved):
Sounds good. Thanks!

PR #1027, opened 2022-10-26
Add `fill` argument to `dataTableOutput()`

cpsievert:
This PR reverts #1026 and instead adds a `fill` argument to `dataTableOutput()` (which gets passed to `htmlwidgets::shinyWidgetOutput()`). 

Since #1026, https://github.com/ramnathv/htmlwidgets/pull/442 got updated with a generally better way to support fill behavior htmlwidgets (it won't require changing the function signature to `height = NULL`), and now, hopefully more obvious how to opt-out of that behavior if it happens to be undesirable (`fill = FALSE`).

Happy to provide more context if you'd like.


NEWS.md
```diff
@@ -1,6 +1,7 @@
 # CHANGES IN DT VERSION 0.27
 
-- `dataTableOutput()`'s `height` and `width` now defaults to `NULL`, allowing it have more flexible default sizing behavior, which will be primarily useful in combination with `{bslib}`'s new `card()` API (thanks, @cpsievert, #1026).
+- `dataTableOutput()` gains a new `fill` parameter. When `TRUE` (the default), the widget's container element is allowed to grow/shrink to fit it's parent container so long as that parent is opinionated about its height and has been marked with `htmltools::bindFillRole(x, container = TRUE)`. (#2198)
```
yihui, line 3:
```suggestion
- `dataTableOutput()` gains a new `fill` parameter. When `TRUE` (the default), the widget's container element is allowed to grow/shrink to fit it's parent container so long as that parent is opinionated about its height and has been marked with `htmltools::bindFillRole(x, container = TRUE)`. (#2198)

```

yihui (approved):
Sounds good. Thanks!

R/shiny.R
```diff
@@ -22,11 +22,13 @@
 #'     }
 #'   )
 #' }
-dataTableOutput = function(outputId, width = NULL, height = NULL) {
+dataTableOutput = function(outputId, width = '100%', height = 'auto', fill = TRUE) {
```
yihui, line 25: The new argument needs to be documented.

yihui (changes requested):

PR #1045, opened 2023-01-08
Fix `doGlobalSearch()` with tibbles

### LINKED ISSUE #1044: `doGlobalSearch()` gives wrong result for tibbles
`doGlobalSearch()` gives an incorrect result for tibble input data frames. This is likely due to the `drop = FALSE` default when subsetting tibbles.

``` r
tbl <- tibble::tibble(x = c("A", "B"))

# Incorrect
DT::doGlobalSearch(tbl, "B")
#> [1] 1

# Expected
DT::doGlobalSearch(as.data.frame(tbl), "B")
#> [1] 2
```

mikmart:
Fixes #1044.

I added tibble as a Suggested package in order to add a regression test. Would there be a better way to keep a regression test, and avoid adding the dependency?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -43,7 +43,8 @@ Suggests:
     shiny (>= 1.6),
     bslib,
     future,
-    testit
+    testit,
+    tibble
 VignetteBuilder: knitr
 RoxygenNote: 7.2.3
 Encoding: UTF-8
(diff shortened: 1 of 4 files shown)


yihui (approved):
Adding soft dependencies is always fine. Thank you!

PR #1049, opened 2023-02-08
Allows custom column filter placeholder

andyquinterom:
This PR allows users to set a custom placeholder for column filters. Currently, it is hard coded to `'All'` which makes it almost impossible to setup easily in other languages. (At our company we use Spanish).

This would be really helpful.

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -107,6 +107,9 @@
 #'   columns and the text areas for some other columns by setting
 #'   \code{editable} to a list of the form \code{list(target = TARGET, numeric
 #'   = INDICES1, area = INDICES2)}.
+#' @param columnFilterPlaceholder a length-one character vector to specify the
+#'   placeholder text for the column filter input boxes. Defaults to
+#'   \code{"All"}.
 #' @details \code{selection}:
 #'   \enumerate{
 #'     \item The argument could be a scalar string, which means the selection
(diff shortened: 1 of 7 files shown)


yihui (commented):
It seems you want to change the language globally (to Spanish), so I think a cleaner approach may be using a global option, e.g., `options(DT.columnFilterPlaceholder = ...)`. That means this PR can be simplified---just replace `'All'` with `getOption('DT.columnFilterPlaceholder', 'All')`. Would you be okay with that?

`datatable()` has already had too many arguments, and I tend not to add more.

andyquinterom:
> It seems you want to change the language globally (to Spanish), so I think a cleaner approach may be using a global option, e.g., `options(DT.columnFilterPlaceholder = ...)`. That means this PR can be simplified---just replace `'All'` with `getOption('DT.columnFilterPlaceholder', 'All')`. Would you be okay with that?
> 
> `datatable()` has already had too many arguments, and I tend not to add more.

Sounds good to me. Shouldnt there be an API that's easier to access for users? since there would be two forms of internationalization, the JSON file passed on to the options.language.url and now this. I understand DataTables is a separate project, but a unified API for Shiny users could be cool to see. I could sit down for a while and think of some sort of unified API for DT internationalization.

What do you think?

yihui:
Yes, a unified API is definitely a good idea. Please feel free to make suggestions. Thanks!

Issue #1052, opened 2023-03-06
DataTable `class` list options don't work with `style = 'bootstrap'`


cpsievert:
```r
# works
DT::datatable(mtcars, class=list(stripe=F))
# doesn't work
DT::datatable(mtcars, class=list(stripe=F), style = "bootstrap")
#> Error in strsplit(class, "\\s+") : non-character argument
```

This happens because `DT2BSClass()` can only handle a character vector...seems like it should be also able to handle a list?

https://github.com/rstudio/DT/blob/c1cdbef524ec3c978cd140a5294d1330c30d06d6/R/datatables.R#L936-L952

PR #1089, opened 2023-09-27
Add support for list of classes in `DT2BSClass`

pedropark99:
This PR seeks to bring support for a list of classes in `DT2BSClass()`, with the end goal of fixing the issue #1052.

In more details, when you select `style='bootstrap'`, `DT::datatable()` tries to translate the `class` argument, to a css rule with all the Bootstrap classes that were mentioned in the `class` argument. To make this translation, the `DT2BSClass()` function is called with the class argument. However, the `DT2BSClass()` function only supports a character vector as input.

This PR will allow `DT2BSClass()` to receive a list of booleans as input, as mentioned in the example from issue #1052.
```r
# Now these examples below works
DT::datatable(mtcars, class=list(stripe=F), style = "bootstrap")
DT::datatable(mtcars, class=list(stripe=T, compact=F, hover=T), style = "bootstrap")
```

In essence, we use `base::Filter()` to select the elements of the list that are TRUE, and simply extract the name of these filtered elements.
```r
  if (is.list(class)) {
    chosen_classes = Filter(function(x) isTRUE(x), class)
    class = names(chosen_classes)
  }
```

### DIFF
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -933,8 +933,7 @@ DTDependencies = function(style) {
 # translate DataTables classes to Bootstrap table classes
 DT2BSClass = function(class) {
   if (is.list(class)) {
-    chosen_classes = Filter(function(x) isTRUE(x), class)
-    class = names(chosen_classes)
+    class = names(which(unlist(class)))
   } else {
     class = unlist(strsplit(class, '\\s+'))
   }
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -938,10 +938,8 @@ DT2BSClass = function(class) {
   } else {
     unlist(strsplit(class, '\\s+'))
   }
-
-  if ('display' %in% class) {
+  if ('display' %in% class)
     class = unique(c('stripe', 'hover', 'row-border', 'order-column', class))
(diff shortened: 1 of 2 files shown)


R/datatables.R
```diff
@@ -932,9 +932,17 @@ DTDependencies = function(style) {
 
 # translate DataTables classes to Bootstrap table classes
 DT2BSClass = function(class) {
-  class = unlist(strsplit(class, '\\s+'))
-  if ('display' %in% class)
+  if (is.list(class)) {
+    chosen_classes = Filter(function(x) isTRUE(x), class)
+    class = names(chosen_classes)
```
yihui, line 937: Perhaps this suffices?

```suggestion
    class = names(which(unlist(class)))
```

yihui (approved):
Looks good. It will be great to add a NEWS item to `NEWS.md`. Thanks!

pedropark99:
Thank you @yihui  for the review! I updated NEWS.md with the changes, and, also changed the source code to use `which()` instead of `Filter()` as you suggested 😉

Issue #1069, opened 2023-06-06
Sorting a reordered column (w/ ColReorder) sorts column at original position


isthisthat:
Not sure whether this is a regression since I can see it working fine [on the demo page](https://rstudio.github.io/DT/extensions.html).
Here's an example where I've swapped the position of two columns and ordered by one (gc_rank), which results in the table being ordered by the other (c_rank), which is the original position of the "gc_rank" column. Having the same issue even when I keep all other settings as default and not loading any other extensions. I'm using DT v0.28 in shiny v1.7.4.
![image](https://github.com/rstudio/DT/assets/1259045/4d47a00e-00b5-4aee-b5ca-cc2f9f87c87f)

```
> xfun::session_info('DT')
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 22.04.1 LTS, RStudio 2022.7.2.576

Locale:
  LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8     LC_MONETARY=en_GB.UTF-8   
  LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
  LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3   bslib_0.4.2       cachem_1.0.8      cli_3.6.1         crosstalk_1.2.0   digest_0.6.31     DT_0.28           ellipsis_0.3.2   
  evaluate_0.21     fastmap_1.1.1     fontawesome_0.5.1 fs_1.6.2          glue_1.6.2        graphics_4.1.2    grDevices_4.1.2   highr_0.10       
  htmltools_0.5.5   htmlwidgets_1.6.2 jquerylib_0.1.4   jsonlite_1.8.4    knitr_1.42        later_1.3.1       lazyeval_0.2.2    lifecycle_1.0.3  
  magrittr_2.0.3    memoise_2.0.1     methods_4.1.2     mime_0.12         promises_1.2.0.1  R6_2.5.1          rappdirs_0.3.3    Rcpp_1.0.10      
[...]

Issue #921, opened 2021-07-09
Shiny and Colreorder, doesn't sort columns properly after drag and drop


nmoraesmunter:
Minimal code to reproduce the issue (shiny v1.6.0 DT v0.18)

```r

library(shiny)
library(DT)

ui <- fluidPage(
  br(),
  DTOutput("tbl"),
  br()
)

server <- function(input, output, session){
  
  output$tbl <- renderDT({
    datatable(mtcars, extensions = "ColReorder", 
              options = list(colReorder = TRUE)
    )
  })
  
}

shinyApp(ui, server)

```

1. Click in a column header (for example qsec). Check that it's sorted correctly. (ascendent order of qsec values 14.5, 14.6, 15.41, 15.5...)
2. Drag the column to another position ( I put it one column to the left, between drat and wt)
3. Click in the column header of the dragged column. Check that it's not sorting that column properly (ascendent order of qsec values 16.9, 18.52, 19.9, 18.9, 16.7..)

It seems to me that it looses the right indexing when dragging and dropping and it is ordering by another column. I have tried to run it without Shiny and it works as expected, that's why I think that the issue is when interacting with Shiny.

```r
datatable(mtcars, extensions = "ColReorder", 
              options = list(colReorder = TRUE)
    )
```

Posted in Stackoverflow https://stackoverflow.com/questions/68302326/shiny-and-colreorder-doesnt-sort-properly-after-column-after-drag-and-drop 

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('DT')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/DT')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

nmoraesmunter:
I deployed the minimal code into https://nmuenter.shinyapps.io/colreorder/ where the issue can be reproduced

nmoraesmunter:
[...]

PR #1096, opened 2023-11-15
Bugfix: Correctly sort re-ordered columns

### LINKED ISSUE #534: minor colReorder problem remaining
#532 is awesome, thanks so much @shrektan ! I did encounter one remaining bug - it only happens with server set to TRUE, so possibly it should just be mentioned in the documentation.

 To reproduce:

1) Create a DT with colReorder, a filter, and with server set to TRUE
2) Use the filter to filter a column
3) Reorder the column that you've just filtered
4) Try to reset the filter

When you do this, it's clear that the filter isn't being applied to the right column. In the example below, I get a javascript popup about the numeric type vs character type, but if all columns are character, then there's no error but I believe the old filter is applied to the wrong column.

```r
# Same a #531 but this problem is only for server=TRUE
library(shiny)
library(DT)

ui <- basicPage(
  DT::dataTableOutput('foo')
)

  output$foo = DT::renderDataTable(
    iris, server = TRUE,
    filter = "top", 
    caption = 'colReorder + filter can have trouble',
    extensions = 'ColReorder',
    options = list(colReorder = TRUE)
  )
})
runApp(shinyApp(ui, server))
```
### Initial State
image

### Set filter
image

### Move filtered column
image

### Try to reset filter
image

ashbaldry:
The re-ordered column order isn't sent server-side, so have added the column names to the DT options. This gets passed into the server-side parameters in `columns$[[col_id]]$name` (adding it to `data` instead caused errors)

This PR should resolve #534 , #921 and #1069

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 - Upgraded DataTables version to 1.13.6 (thanks, @stla, #1091).
 
+- Searching and sorting work now when columns are re-ordered by the `ColReorder` extension (thanks, @ashbaldry #1096, @gergness #534, @nmoraesmunter #921, @isthisthat #1069).
+
 - Fixed disabling selection on hyperlink clicks (thanks, @guoci, #1093).
 
 - Fixed an error for R >= 4.3.0 (thanks, @AntoineMichelet, #1095).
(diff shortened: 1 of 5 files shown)


yihui (approved):
Clever! I didn't know that column names would be sent to the server side. Thank you so much! I'll tweak the PR a little bit and merge.

yihui:
It turns out that sorting is easier to fix, and it took me quite a while to figure out how to fix searching, thanks to your guidance.

I think this PR is ready to go now.

ashbaldry:
Glad I could help! Didn't realise that the search was also not working when re-ordered

PR #1111, opened 2024-01-12
Update numeric slider values on updateSearch

### LINKED ISSUE #1110: updateSearch does not update numeric slider values
`updateSearch()` does not update the slider values according to the newly set search string for numeric columns.

``` r
library(shiny)
library(DT)

ui <- fluidPage(
  actionButton("set_search", "Set column search"),
  tags$hr(),
  DTOutput("iris")
)

  output$iris <- renderDT(
    datatable(iris, filter = "top")
  )

  proxy <- dataTableProxy("iris")
  observeEvent(input$set_search, {
    updateSearch(proxy, keywords = list(columns = c("", "5.0 ... 6.0")))
  })
}

shinyApp(ui, server)
```

image

```
> xfun::session_info('DT')
R version 4.3.2 (2023-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19045), RStudio 2023.12.0.369

Locale:
  LC_COLLATE=English_United Kingdom.utf8  LC_CTYPE=English_United Kingdom.utf8   
  LC_MONETARY=English_United Kingdom.utf8 LC_NUMERIC=C                           
  LC_TIME=English_United Kingdom.utf8    

Package version:
  base64enc_0.1.3   bslib_0.6.1       cachem_1.0.8      cli_3.6.2         crosstalk_1.2.1  
  digest_0.6.34     DT_0.31.1         ellipsis_0.3.2    evaluate_0.23     fastmap_1.1.1    
  fontawesome_0.5.2 fs_1.6.3          glue_1.7.0        graphics_4.3.2    grDevices_4.3.2  
  highr_0.10        htmltools_0.5.7   htmlwidgets_1.6.4 httpuv_1.6.13     jquerylib_0.1.4  
  jsonlite_1.8.8    knitr_1.45        later_1.3.2       lazyeval_0.2.2    lifecycle_1.0.4  
  magrittr_2.0.3    memoise_2.0.1     methods_4.3.2     mime_0.12         promises_1.2.1   
  R6_2.5.1          rappdirs_0.3.3    Rcpp_1.0.12       rlang_1.1.3       rmarkdown_2.25   
  sass_0.4.8        stats_4.3.2       stringi_1.8.3     stringr_1.5.1     tinytex_0.49     
  tools_4.3.2       utils_4.3.2       vctrs_0.6.5       xfun_0.41         yaml_2.3.8  
```

---

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

mikmart:
Fixes #1110.

``` r
library(shiny)
library(DT)

ui <- fluidPage(
  actionButton("set_search", "Set column search"),
  tags$hr(),
  DTOutput("iris")
)

  output$iris <- renderDT(
    datatable(iris, filter = "top")
  )

  proxy <- dataTableProxy("iris")
  observeEvent(input$set_search, {
    updateSearch(proxy, keywords = list(columns = c("", "5.0 ... 6.0")))
  })
}

shinyApp(ui, server)
```

image

### DIFF
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -1402,9 +1402,9 @@ HTMLWidgets.widget({
           console.log('The search keyword for column ' + i + ' is undefined')
           return;
         }
-        // Change event is needed to update numeric slider values.
+        // Update column search string and values on linked filter widgets.
+        // 'input' for factor and char filters, 'change' for numeric filters.
         $(td).find('input').first().val(v).trigger('input').trigger('change');
-        searchColumn(i, v);
       });
       table.draw();
     }
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -498,9 +498,6 @@ HTMLWidgets.widget({
               var v1 = JSON.stringify(filter[0].selectize.getValue()), v2 = $input.val();
               if (v1 === '[]') v1 = '';
               if (v1 !== v2) filter[0].selectize.setValue(v2 === '' ? [] : JSON.parse(v2));
-            },
-            updatesearch: function() {
-              $input.trigger('input');
             }
           });
           var $input2 = $x.children('select');
@@ -536,13 +533,7 @@ HTMLWidgets.widget({
           if (server) {
             fun = $.fn.dataTable.util.throttle(fun, options.searchDelay);
           }
-          $input.on({
-            input: fun,
-            updatesearch: function() {
-              // Bypass any throttling.
-              searchColumn(i, $input.val()).draw();
-            }
-          });
+          $input.on('input', fun);
         } else if (inArray(type, ['number', 'integer', 'date', 'time'])) {
           var $x0 = $x;
           $x = $x0.children('div').first();
@@ -626,10 +617,6 @@ HTMLWidgets.widget({
               if (v[0] != slider_min()) v[0] *= scale;
               if (v[1] != slider_max()) v[1] *= scale;
               filter.val(v);
-            },
-            updatesearch: function () {
-              $input.trigger('input'); // Resets slider if search is empty.
-              $input.trigger('change');
             }
           });
           var formatDate = function(d, isoFmt) {
@@ -1415,7 +1402,9 @@ HTMLWidgets.widget({
           console.log('The search keyword for column ' + i + ' is undefined')
           return;
         }
-        $(td).find('input').first().val(v).trigger('updatesearch');
+        // Update column search string and values on linked filter widgets.
+        // 'input' for factor and char filters, 'change' for numeric filters.
+        $(td).find('input').first().val(v).trigger('input').trigger('change');
       });
       table.draw();
     }
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -530,10 +530,16 @@ HTMLWidgets.widget({
           var fun = function() {
             searchColumn(i, $input.val()).draw();
           };
-          if (server) {
-            fun = $.fn.dataTable.util.throttle(fun, options.searchDelay);
-          }
-          $input.on('input', fun);
+          var throttledFun = $.fn.dataTable.util.throttle(fun, options.searchDelay);
+          $input.on('input', function(e, immediate) {
+            if (immediate) {
+              fun(); // Allows updateSearch to bypass throttling.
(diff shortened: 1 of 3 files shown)


inst/htmlwidgets/datatables.js
```diff
@@ -1402,7 +1402,8 @@ HTMLWidgets.widget({
           console.log('The search keyword for column ' + i + ' is undefined')
           return;
         }
-        $(td).find('input').first().val(v).trigger('input');
+        // Change event is needed to update numeric slider values.
+        $(td).find('input').first().val(v).trigger('input').trigger('change');
```
yihui, line 1406: If we trigger the `change` event, it seems `searchColumn()` on the next line is no longer necessary. If that sounds correct to you, would you help test it?

yihui (approved):
Thank you very much!

inst/htmlwidgets/datatables.js
mikmart: Yep it looks like it will get triggered by the events anyway. Quickly testing removing it, the numeric filter in the example here seems to keep working fine. I'm happy to have a more thorough look at the other filter types tomorrow.

mikmart: I tested removing it with all 3 filter types now, using the app below. As far as I can tell, everything keeps working as intended. I also logged the calls to `searchColumn()` and they happen as intended. I pushed the change.

``` r
library(shiny)
library(DT)

tbl <- data.frame(
  num1 = seq(-2, 2) / 10,
  fct1 = factor(c("A", rep("B", 3), "C")),
  chr = c("foo", "bar", "baz", "baz", "baz")
)

ui <- fluidPage(
  DTOutput("table"),
  actionButton("update_search", "Update search"),
  actionButton("clear_search", "Clear search"),
)

  output$table <- renderDT(datatable(tbl, filter = "top", rownames = FALSE))
  
  observeEvent(input$update_search, {
    updateSearch(proxy, list(columns = c("-0.1 ... 0.1", '["B"]', 'bar')))
  })
  observeEvent(input$clear_search, clearSearch(proxy))
}

shinyApp(ui, server)
```

mikmart: Although now I do notice that sometimes when updating multiple search strings at a time, the search based on the numeric column happens first, and it takes a few moments for the other to kick in. That seems undesirable. Probably related to always `.draw()`ing when searching.

mikmart: Okay the delay is because of the character search field input throttling and only happens because when testing I'm flicking quickly between "update" and "clear". It's made apparent because the numeric search triggers the `.draw()` early. Not sure why the factor filter wouldn't do the same even without this patch, though.
https://github.com/rstudio/DT/blob/487dc08fe757fe0be68931c8b88395f11a9c155a/inst/htmlwidgets/datatables.js#L533-L535

mikmart: Oh I see it's because the removed `searchColumn()` was bypassing the throttling on the character search.

mikmart:
I changed the approach to use a custom event `updatesearch` so that each search field can handle updates from R as it wishes. This was to avoid the unnecessary `searchColumn()` call while also bypassing the input throttling for character columns.

Would you have another look @yihui? In particular I'm wondering if the custom event name should be namespaced somehow.

yihui (approved):
I thought for a while, and perhaps we can pass an extra parameter to the `input` event handler to decide whether to throttle the search. You can find jQuery's `trigger()` documentation here: https://api.jquery.com/trigger/ The idea is that our handler `fun` takes two arguments `e` (event) and `immediate` (whether to search immediately or throttle the search). `immediate` is `false` by default, and we set it to `true` in the `updateSearch()` method, e.g.,

```js
$(td).find('input').first().val(v).trigger('input', [true]);
```

I feel that might be a little simpler than defining an `updatesearch` event for all types of inputs. Let me know if the idea is not clear to you. I can also tweak the PR by myself. Thanks!

mikmart:
Yeah that makes sense. I can pivot to that approach.

mikmart:
Ready for a review again @yihui.

yihui (approved):
Looks perfect now! I'll make a cosmetic change and merge soon. Thank you so much!

PR #1113, opened 2024-01-20
Add server-side processing support for SearchBuilder

### LINKED ISSUE #963: [FR] SearchBuilder works with Server-Side processing mode
Hello, so a really helpful extension added to this repo was SearchBuilder, and while using it, once i go to server processing mode (server = true), it doesn't work, and this is a must on my side because i am dealing with fairly large data,

So is there any solution for this?, whether to optimize the data flow on client side (server = false) or to make the extension work on server side processing?

Regards,

---

- [ ] I have fully read the issue guide at https://yihui.name/issue/.
- [ ] I have provided the necessary information about my issue.
- [ ] I have learned the Github Markdown syntax, and formatted my issue correctly.

mikmart:
Fixes #963.

This PR adds support for SearchBuilder in server-side processing. I did not add support for the `moment` or `luxon` column types as I'm not sure how they map to R types. The `array` column type is not supported either.

Notably datetime columns are treated as character data. This seems to be a limitation with the SearchBuilder extension.

I was not sure how to map the `null` or empty conditions into R. The approach I chose was to map them to missing values, and, for character columns, also to empty strings.

I added some unit tests to confirm that the main components in the processing work as intended, but did not attempt to get thorough coverage on all cases. In addition to the unit tests, I used this app for some further manual testing:

``` r
library(shiny)

shinyApp(
  fluidPage(DT::DTOutput('foo')),
  function(input, output) {
    output$foo = DT::renderDT(
      data.frame(
        a = sample(26), b = letters,
        c = factor(rep(c('a', 'b'), 13)),
        d = Sys.Date() + 1:26,
        e = Sys.time() + 1000 * (1:26)
      ),
      options = list(dom = 'Qlfrtip'),
      extensions = c('SearchBuilder', 'DateTime')
    )
  }
)
```


yihui (approved):
Wow! This is incredible. Thank you so much!

Please add a news item and I think it'll be ready to merge.

mikmart:
Happy to help. And thank you for blogging about this! You were right, this _was_ a fun project.

I added the NEWS item and also:

1. Verified that both empty and `NA` strings count for `null` in client-side processing.
2. Encoded the string `null` evaluation rules in a unit test.
3. Removed the second logic evaluation test that was, on second thought, a bit redundant.

AhmedKhaled945:
Thanks a lot for the work @mikmart , @yihui 

Just a question that i cannot find an answer to, i used the code above to test, works like a charm, but for the factor column (c), when i try to add a condition, it is giving me a textbox instead of dropdown of choices, is this how it is supposed to be? or should i add another option somewhere to let me choose from fixed values (since it is a factor),

image


Thanks in advance,

mikmart:
Thanks for trying it out @AhmedKhaled945! What you're seeing is actually a [documented feature](https://datatables.net/extensions/searchbuilder/serverside) of the SearchBuilder extension in server-side processing mode (emphasis added):

> There are two caveats that SearchBuilder's server-side integration carries. The first is that **anywhere select elements would normally be used on the client-side, input elements are used instead**. This reduces strain on the server significantly, drastically improving performance.

I'm not aware of any straightforward way to change this behaviour.

AhmedKhaled945:
Ok got it, thanks a lot @mikmart

PR #1114, opened 2024-01-21
Ignore invalid criteria in SearchBuilder server-side processing

mikmart:
Follow-up to #1113.

Sometimes the client will send SearchBuilder queries that contain incomplete or invalid search criteria. For example, when the user changes the `data` that an existing criteria applies to, a search will be sent to the server with the `condition` component missing. In client-side processing, such invalid criteria are ignored.

This PR aligns server-side processing with the client-side implementation, by returning `NULL` to signal that a condition is invalid and should be ignored or that the search contains no valid criteria.

### DIFF
--- a/R/searchbuilder.R
+++ b/R/searchbuilder.R
@@ -1,12 +1,13 @@
 # server-side processing for the SearchBuilder extension
 # https://datatables.net/extensions/searchbuilder/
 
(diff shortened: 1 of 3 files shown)


yihui (approved):
Thanks!

PR #1119, opened 2024-01-24
Apply column formats to slider labels

### LINKED ISSUE #247: Is it possible to have the column formatting also apply to values shown in filter sliders?
For example, I can use "formatPercentage" to format the values in a column to look like percentages (e.g. 2.50% (with "digits = 2") instead of 0.025), but the values shown on the filter slider still look unformatted (e.g. 0.025 instead of 2.50%).

mikmart:
Fixes #247.

The `format*` functions append new `columnDefs` entries with customized renderer functions to apply column formatting. The idea in this PR is to fetch those same renderer functions to apply to filter slider labels.

``` r
DT::datatable(
  tibble::tibble(
    goal_completion = round(runif(10), 2),
    bonus = goal_completion * 15:24 * 1000,
    payday = Sys.Date() + 1:10
  ),
  filter = "top"
) |>
  DT::formatPercentage(1) |>
  DT::formatCurrency(2) |>
  DT::formatDate(3)
```
rstudio_q1oywttPsw

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 - Fixed a bug that caused the date picker widget from the DateTime extension to render incorrectly (thanks, @mikmart, #1116).
 
+- Column formatting now also applies to range labels shown on filter sliders (thanks, @GitChub, @mikmart, #247).
+
 # CHANGES IN DT VERSION 0.31
 
 - Upgraded DataTables version to 1.13.6 (thanks, @stla, #1091).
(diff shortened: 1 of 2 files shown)


yihui (approved):
Implementing a feature requested from eight (!) years ago... How awesome is that! Plus, the implementation is much more elegant than what I have thought before. Thank you so much!

mikmart:
Thank you for the kind words!

I think with these changes the previously special-cased date formatting can also be simplified. I'm planning to have a stab at refactoring that in another PT.

PR #1122, opened 2024-01-25
Remove special-casing of dates for slider formatting and fix datetimes

mikmart:
Follow-up to #1119.

This PR removes the now-unnecessary special-casing from `formatDate()` as the mechanism from #1119 handles it directly.

This also fixes a regression introduced in #1119 for datetime columns, which was caused by me overlooking to handle datetime values when applying a renderer to format the slider range labels.

There's some duplicated logic in the JS side `formatDate()` helper and `restore()`. I thought about refactoring that, but it would be a slightly more involved operation without a big direct benefit, so I decided not to mess with it at this point.

I tested manually with this app:

``` r
DT::datatable(
  data.frame(
    date = Sys.Date() + 1:10,
    time = Sys.time() + 1:10 * 60 * 60 * 24
  ),
  filter = "top"
) |> DT::formatDate(1:2, "toLocaleString")
```

### DIFF
--- a/R/format.R
+++ b/R/format.R
@@ -127,8 +127,6 @@ formatSignif = function(
 #'   \code{params = list('ko-KR', list(year = 'numeric', month = 'long', day =
 #'   'numeric'))}
 formatDate = function(table, columns, method = 'toDateString', params = NULL, rows = NULL) {
-  if (!inherits(table, 'datatables'))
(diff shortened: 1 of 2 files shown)


yihui (approved):
Excellent! Thanks!

PR #1126, opened 2024-02-26
fix selection sync on client mode update (fixes #1125)

### LINKED ISSUE #1125: #828 breaks maintaining selected columns upon re-render
https://github.com/rstudio/DT/blob/06d3ae65b5e9b1b0c494de8cacaccfc60c7018cc/inst/htmlwidgets/datatables.js#L995-L1001

The above code was added in response to #828, where the server side was not cleared when the table is reloaded and the Select extension is in use. However, this code doesn't _sync_, it _wipes_. So if user code, e.g. via callback, attempts to _retain_ the row selections made by the user, only updating other parts of the table, the server side gets out of sync again. And since this piece of code is executed after the callbacks, it overrides any `Shiny.setInputValue()` that might be used to correct the server side. 

Example:
```
DT::datatable(
  callback = DT::JS(
    paste0("var rows = [", rows_str, "];"),
    paste0("var reactvar = '", id, "_rows_selected';"),
    "table.rows(rows).select();",
    "Shiny.setInputValue(reactvar, rows);"
  )
)
```

The above sets the selected rows according to the numbers in `row_str`, e.g. `0,1,2`, then would send the same back to Shiny (yes, probably off by one due to R's 1-indexing). However, the message is overwritten by the mentioned code.

I'd suggest removing the `cleanSelectedValues()` call entirely, and instead calling the `updateSelectedRows()` family of functions defined directly below. That way, R get's a copy of the actual status.

If anyone has a suggestion for a workaround, I'm all ears... Perhaps having a hook towards the end of the `renderValue` function would be helpful, to be able to override it when it does something other than necessary in a specific case.

epruesse:
This is "minimally invasive". After the reload of the table on the client and potentially restoring selections via JS callback function, the `cleanSelectedValues()` function sends potentially incorrect `<tbl>_xxx_selected` information to R. The patch uses the functions already implemented to send row/col/cell selections made via Select extension to R. It simply triggers one update right then and there. Since the messages get merged, this effectively overwrites the false message created by cleanSelectedValues() exactly when the bug #1125 is triggered. It may not be a complete fix for other scenarios, but it is brief and has minimal chance of breaking things for other use cases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - Fixed a bug in `renderDT()`'s evaluation of `...` arguments when `quoted = TRUE` (#1130). 
 
+- Maintain selected columns/rows/cells upon re-render (thanks, @epruesse, #1125).
+
 # CHANGES IN DT VERSION 0.32
 
 - Fixed a bug that caused the column used for grouping with the RowGroup extension to change when relocated by the ColReorder extension (thanks, @isthisthat, @mikmart, #1109).
--- a/inst/htmlwidgets/datatables.js
+++ b/inst/htmlwidgets/datatables.js
@@ -1032,6 +1032,9 @@ HTMLWidgets.widget({
         updateColsSelected();
         updateCellsSelected();
       })
+      updateRowsSelected();
+      updateColsSelected();
+      updateCellsSelected();
     }
 
     var selMode = data.selection.mode, selTarget = data.selection.target;


yihui (approved):
Thanks! This sounds about right. Please add a news item to `NEWS.md` under v0.33 and I'll merge this PR.

PR #1157, opened 2024-11-25
Add lazyRender parameter

### LINKED ISSUE #1156: Render table before it is displayed
An issue that I keep running into quite frequently is that I need to perform proxy actions on a table in a shiny app before it ever becomes visible.  This is usually when the table is in a conditional panel whose condition has never been true, or if the table is in a tab within a `tabsetPanel` that has never been active.

Interestingly, the `replaceData` action will work if it is run before the table is ever rendered, and when it is rendered it will show the correct data, but most, if not all other proxy actions will not work until the table has been rendered first.

Here is a demo app.  The table is on the second tab and will not render until the second tab is activated.  On the first tab are two buttons.  The "Select 1st Row" button will select the first row of the table.  If that button is clicked before ever activating the second tab, the first row will not be selected when going to the second tab.  If it is clicked after activating the second tab, it will be selected.  The "Replace Data" button will replace the data in the table, and that does work if clicked before ever activating the second tab.

```
library(shiny)
library(DT)

ui <- fluidPage(
  tabsetPanel(
    id = "tabs",
    tabPanel("Tab 1",
             actionButton("select", "Select 1st Row"),
             actionButton("replace", "Replace Data")),
    tabPanel("Tab 2",
             DTOutput("mt_table"))
  )
)

  output$mt_table <- renderDT(mtcars)
  outputOptions(output, "mt_table", suspendWhenHidden = FALSE)
  table_proxy <- dataTableProxy("mt_table")
  
  observeEvent(input$select, {
    
    selectRows(proxy = table_proxy, selected = 1)
    
  })
  
  observeEvent(input$replace, {
    
    replaceData(proxy = table_proxy, data = rbind(mtcars, mtcars))
    
  })
  
}

shinyApp(ui, server)
```

It would be useful if all proxy actions would work even if the table has never been rendered.  Even better would be if there was an option to render the table immediately rather than wait for the table to first become visible.  This would allow for advanced modification of the table beyond just the provided proxy functions.

The only discussion I could find on this was from 2016 [here](https://groups.google.com/g/shiny-discuss/c/4E61F-tvGR4).  It does give a workaround which worked for that scenario.  But in that case the person was wrapping the table within their own div.  The workaround doesn't help when the table is in a conditional panel or a tab.

The explanation given is that due to this part, the table doesn't render until it becomes visible, i.e. if either `offsetWidth` or `offsetHeight` are 0.
https://github.com/rstudio/DT/blob/main/inst/htmlwidgets/datatables.js#L183-L186

It would be useful if a parameter could be passed through `datatable` that could indicate not to delay rendering the table.  Call it something like `lazyRender`.  Then in `datatables.js`, something like:

```
if ((el.offsetWidth === 0 || el.offsetHeight === 0) & data.lazyRender)
```

Admittedly I don't know enough to know if it's really as simple as that or if there's any other considerations to doing this.  If not, perhaps there's some other way to achieve this.

Mosk915:
This resolves #1156.  It adds the option to have the table render immediately rather than waiting for it to become visible.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # CHANGES IN DT VERSION 0.34
 
+- Added `lazyRender` parameter to `datatable`, which gives the option for the table to be rendered immediately rather than waiting for it to become visible (#1156).
 
 # CHANGES IN DT VERSION 0.33
 
--- a/R/datatables.R
+++ b/R/datatables.R
@@ -64,6 +64,8 @@
 #'   (only display the table body) when the number of total records is less
 #'   than the page size. Note, it only works on the client-side processing mode
 #'   and the `pageLength` option should be provided explicitly.
+#' @param lazyRender \code{TRUE} to delay rendering until the table becomes visible
+#'   (the default) or \code{FALSE} to render the table immediately on page load.
 #' @param selection the row/column selection mode (single or multiple selection
 #'   or disable selection) when a table widget is rendered in a Shiny app;
 #'   alternatively, you can use a list of the form \code{list(mode = 'multiple',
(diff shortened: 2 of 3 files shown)


Mosk915:
Issue with this pull request.  Will be resubmitting.

